PR #10, opened 2020-08-05
VMO-2224 fixes after resolving conflicts from Brett's styling

safydy-r:
There are few issues which were a bit complicated to me, I'd like to ask you @jacob-viamo to look at them and try to find a solution if possible:
1- Cannot add select many: when adding SelectMany block from menu, SelectOne is added instead
2- Resource Editor not working for new added block
3- When adding CaseBlock, we have no label on exits

### DIFF
--- a/builder.config.json
+++ b/builder.config.json
@@ -12,7 +12,7 @@
     },
     "languages": [
       {
-        "id": "1",
+        "id": "22",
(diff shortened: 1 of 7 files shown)


safydy-r:
found an issue for Resource Editor when adding new block
image

PR #12, opened 2020-10-08
Vmo 2642 move storybook stories from voto5

safydy-r:
Key points:
- Copied from voto5
- Fixed importations & broken variable names
- Loaded `trees` store to be used in stories: useful for localization as well
- Removed the fake localization utils
- Made some reorganization on stories: refactored codes, fixed IVR resource, etc
- Had to fix the resource display
- Tried on local machine & everything is working except Resources on LogBlock, I think we can leave if for now and wait for https://viamoinc.atlassian.net/browse/VMO-2671


src/store/flow/resource.ts
```diff
@@ -222,6 +222,5 @@ export function discoverContentTypesFor(mode: SupportedMode, resource?: IResourc
     }, contentTypeOverrides)
     return contentTypeOverrides
   }, contentTypeOverrides)
-
-  return Object.assign(defaultModeMappings, contentTypeOverrides)[mode]
+  return Object.assign(defaultModeMappings, contentTypeOverrides)[mode][0]
```
safydy-r, line 225: this avoids us to get array inside array values like `"sms":[["text"]]`, and give `"sms":["text"]`

safydy-r: @bzabos  , for info: this seems fixing the resources issue for **new added blocks** (except for LogBlock). However, existing blocks (created from `2019-10-10-shortcut-flow.json`) have resources broken. Maybe a clue for https://viamoinc.atlassian.net/browse/VMO-2671

src/store/index.ts
```diff
@@ -1,17 +1,19 @@
 import {StoreOptions} from 'vuex'
 import {IFlowsState, store as flow} from './flow'
 import {IBuilderState, store as builder} from './builder'
-
+import trees from './trees/trees'
 
 export interface IRootState {
   builder: IBuilderState,
   flow: IFlowsState,
+  trees: any
```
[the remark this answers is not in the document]
safydy-r: The trees store is loaded from InteractionDesigner : https://github.com/FLOIP/flow-builder/blob/master/src/views/InteractionDesigner.vue#L167, not only for stories. Therefore, a block editor component loaded outside this InteractionDesigner will be broken. That's why I tried to add this.
I have a suggestion: What if instead of loading the `trees` store from InteractionDesigner, we put it here (as I added)? (meaning, I need to change the store import in InteractionDesigner to make things uniform)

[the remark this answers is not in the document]
safydy-r: Implemented this:
- remove some imports on InteractionDesigner: https://github.com/FLOIP/flow-builder/pull/12/files#diff-4778731e3debd774a6c14025179b9bfb34aae3d45fb42448205377ca6475503dL59
- then import the main `store` : https://github.com/FLOIP/flow-builder/pull/12/files#diff-4778731e3debd774a6c14025179b9bfb34aae3d45fb42448205377ca6475503dR57

- And change the module definition here: https://github.com/FLOIP/flow-builder/pull/12/files#diff-4778731e3debd774a6c14025179b9bfb34aae3d45fb42448205377ca6475503dR163

What do you think ?

src/stories/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+var eslintConfig = require('../.eslintrc.js');
+
+module.exports = Object.assign({}, eslintConfig, {
+  "rules": Object.assign({}, eslintConfig.rules, {
+    //necessary to allow devDependencies required for storybook
+    //in future this could be restricted to only the stories folder
```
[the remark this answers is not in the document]
safydy-r: Ok, trying to find how to restrict this ;)

src/store/index.ts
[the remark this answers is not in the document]
jcbashdown: I think this is a definite improvement. @bzabos Do you forsee any problems with this?

src/stories/.eslintrc.js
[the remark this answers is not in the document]
safydy-r: @jcbashdown , I tried this and it is already restricted to current folder only (which is `stories/` and its subfolders)

Eg. this is an error when I import a devDependencies within a file which is outside `stories/` (but I don't have the error if the file is inside `stories/`)
Screen Shot 2020-10-22 at 7 51 43 PM

So what I only removed this comment to not lead us to confusion. Any concerns ?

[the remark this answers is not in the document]
jcbashdown: Great! Thanks for clearing that up

src/store/flow/resource.ts
safydy-r: I removed this change in this current PR

src/store/flow/resource.ts
```diff
@@ -143,15 +143,17 @@ export function findResourceVariantOverModesWith(
...
-              && difference(filter.modes, v.modes).length === 0)
-
+      filterWithComparatorKeys = pick(filter, keysForComparison) //eg: {"languageId":"1","contentType":"text"}
+  let filterWithComparatorKeysAndArrayContentType = filterWithComparatorKeys //eg: {"languageId":"1","contentType":["text"]} //TODO: we may need to remove this one, once we have a clear idea which schema is used
+  if(!Array.isArray(filterWithComparatorKeysAndArrayContentType.contentType)) {
+    // @ts-ignore
+    filterWithComparatorKeysAndArrayContentType.contentType = [filterWithComparatorKeysAndArrayContentType.contentType]
+  }
+  let variant = find<IResourceDefinitionVariantOverModes>(
+          resource.values, v => (isEqual(filterWithComparatorKeys, pick(v, keysForComparison)) || isEqual(filterWithComparatorKeysAndArrayContentType, pick(v, keysForComparison)))
```
jcbashdown, line 155: If I understand this correctly, could we just use filterWithComparatorKeysAndArrayContentType? Aren't they always going to be equal?

safydy-r: @jcbashdown , thanks for pointing this out. You know what? I think this is a `block resource` fix attempt I made too. And as discussed in previous comment (https://github.com/FLOIP/flow-builder/pull/12#discussion_r501989526), let's avoid such fix in this current PR. 
I'll revert to the original state.

safydy-r: reverted !

src/stories/MessageBlock.stories.ts
```diff
@@ -0,0 +1,124 @@
...
+        languages: {
+          0: {id: languageId}
+        },
+      }: IFlow = this.activeFlow
+      const resourceId = get(this.activeBlock, `config.prompt`, '')
+
+      const variantSms: IResourceDefinitionVariantOverModesFilter = {
+        languageId,
+        modes: [SupportedMode.SMS],
+        contentType: SupportedContentType.TEXT,
```
jcbashdown, line 77: Would it be enough to just wrap contentType in an array here?

safydy-r: resolved as discussed
- used an array `[contentType: SupportedContentType.TEXT]` instead
- then added `ts-ignore` & `TODO comment`

jcbashdown (approved):
I think we can now merge

jcbashdown (approved):

src/stories/story-utils/storeSetup.ts
```diff
@@ -3,23 +3,37 @@ import Vue from 'vue'
...
+  this.$store.registerModule(['flow', BLOCK_TYPE], blockTypeStore)
+}
+
+export const baseMounted = async function (this: any, BLOCK_TYPE: string, blockTypeStore: IRootState): Promise<any> {
+  await safeRegisterBlockModule.bind(this)(BLOCK_TYPE, blockTypeStore)
+
+  const firstFlowId = get(this.$store.state, 'flow.firstFlowId')
+  let flow
+  if (firstFlowId) {
+    flow = findFlowWith(firstFlowId, get(this.$store.state, 'flow') as unknown as IContext)
```
safydy-r, line 30: had to use the same flow if it already exists

jcbashdown: Can you explain a bit more about why this change to base mounted around finding firstFlowId is necessary? Wouldn't it be enough to just add the safeRegisterModel function?

jcbashdown (changes requested):

src/stories/story-utils/storeSetup.ts
safydy-r: I found that in stories we want to have `non starting block`, we fake a case block as activated block. For that, we [use `flow_addBlankBlockByType` to add the case block](https://github.com/FLOIP/flow-builder/pull/12/files#diff-adcd1d3800c16e71c7dcad6f8fbce2afbaf7b6848fce9d877e6eb4241ce59c95R95), but this function doesn't add the case block in the same flow as the first block (non activated) but in the flow mentioned as `firstFlowId`. In fact, we cannot specify the destination `flowId` under `flow_addBlankBlockByType` for now

src/stories/story-utils/storeSetup.ts
```diff
@@ -4,15 +4,17 @@ import {namespace} from 'vuex-class'
...
+let storyInitState: any = {}
 
 /**
  * Safe register block module
  * Because some weird race condition is leading to modules not getting unregistered when clicking between stories before the next story re-registers
  */
 export const safeRegisterBlockModule = async function(BLOCK_TYPE: string, blockTypeStore: IRootState) {
   // @ts-ignore - TS2551: Property '$store' does not exist on type
-  if (this.$store.state['flow'][BLOCK_TYPE]) {
+  if (this.$store.hasModule(['flow', BLOCK_TYPE])) {
```
safydy-r, line 17: this is the correct way to check if a module has been registered, and avoids us having this issue `[vuex] trying to unregister module '....', which is not registered`

jcbashdown (approved):
Looks good!

jcbashdown (approved):

safydy-r:
Approved by @Adea on Nov 16, she has changed the status to `Due for release` on Jira

PR #15, opened 2020-10-14
Vmo 2671 fix misc bugs before qa

bzabos:
Hi @jcbashdown I shared that I'd have this out, but there are two fixes remaining as you review.

- [ ] localize "Choice" prefix and "Log Message" heading
- [ ] fix when we autogen an exit on case block

---

Screen Shot 2020-10-14 at 4 41 19 PM

---

Screen Shot 2020-10-14 at 4 46 55 PM

### DIFF
--- a/builder.config.json
+++ b/builder.config.json
@@ -105,6 +105,7 @@
         "name": "MobilePrimitives\\SelectManyResponse",
         "type": "MobilePrimitives\\SelectManyResponse",
         "is_interactive": true,
+        "is_branching": true,
         "category":  2
       },
       "SmartDevices\\LocationResponse": {
--- a/src/components/interaction-designer/Block.vue
+++ b/src/components/interaction-designer/Block.vue
@@ -52,7 +52,7 @@
           <span class="label label-primary tree-block-item-label tree-block-item-output-subscribers-1"></span>
         </div>
 
-        <h3 class="block-exit-tag label label-warning">{{exit.tag}}</h3>
+        <h3 class="block-exit-tag label label-warning">{{exit.tag || '—'}}</h3>
 
         <template v-if="exit.destinationBlock == null">
           <plain-draggable class="handle-create-link btn btn-default btn-xs btn-flat"
@@ -407,7 +407,7 @@
     }
 
     .block-target {
-      min-height: 6em;
+      min-height: 4em;
       border: 1px dashed transparent;
       border-bottom: 1px solid #eee;
       padding: 0.1em;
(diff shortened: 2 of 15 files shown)


builder.config.json
```diff
@@ -105,6 +105,7 @@
         "name": "MobilePrimitives\\SelectManyResponse",
         "type": "MobilePrimitives\\SelectManyResponse",
         "is_interactive": true,
+        "is_branching": true,
```
jcbashdown, line 108: Not sure if this is how this worked already but I notice we have an exit for each choice with select many when in voto5 we just have a single output for this block. Does it make sense to have an exit per choice when you can make more than one choice?

jcbashdown: I also notice that the example select many at trees/1/edit starts without an empty choice - it has "Yes", "No" and nothing else. I'm not sure if this was always broken but I do know that if I type in "No" then I get an empty choice but then the browser tab freezes and crashes...

jcbashdown (dismissed):
I think this approved just this one commit?

jcbashdown:
I think I'm done with my initial review of this. Are these two still outstanding to be added?

- localize "Choice" prefix and "Log Message" heading
- fix when we autogen an exit on case block

builder.config.json
bzabos: Nice. I'll take a peek and let you know what I find.

bzabos: This does freeze the tab on my local as well. I have done a profile dump of this action, and so have a lead :)

bzabos: Will be fixing + merging this today 👌

bzabos: Hi @jcbashdown — The infinite loop was due to the sample data. 

The deeper issue was that we're doing some data munging to coerce our choices into a list of numeric indices, but I believe these should be allowed to be variable string keys. Combined with the fact that I ripped out some of your code that was filtering resources based upon what existed/didn't exist in order to generate one with `editSelectOneResponseBlockChoice`, I've now left us in a state where we double up on resources. Dang! :) 

What I'm going to do (later this week once I'm back) is update the resource editor behaviour to simply emit an event that contains the resource when it's created to provide a hook to link it and perform any post-op behaviour. I'm still not convinced this is best, though— I'll leet it steep on the back burner for now..

bzabos:
@jcbashdown 
> I think I'm done with my initial review of this. Are these two still outstanding to be added?
> 
> * localize "Choice" prefix and "Log Message" heading
> * fix when we autogen an exit on case block

Yes, these are still outstanding, I will just have to come back to this in a couple days. I have merged so that QA can get on with this.

PR #16, opened 2020-10-22
Vmo 2943 builder as package

jcbashdown:
This is still WIP so you can ignore most of this for now as it's largely linting fixes to make the production build work (and turning off strict). The main things I could use some help with are the changes to package.json and vue.config.js. Note for vue.config.js - I've thrown a bunch of things at the wall to see what sticks in order to turn off chunking. I plan to remove some of this as I suspect it's not all necessary.


bzabos (changes requested):

src/main.ts
```diff
@@ -12,3 +15,8 @@ new Vue({
   store: new Vuex.Store({}),
   render: (h) => h(App),
 }).$mount('#app')
+
+// required inline due to front-loading of imports and having jQuery dependency
+window.onload = function() {
+  require('bootstrap')
```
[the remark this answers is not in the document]
jcbashdown: Currently we import this at the top of bootstrap-legacy-global-dependencies.js and then do some funky stuff with exported. I'm not too sure how to combine this suggestion with that.

src/store/trees/trees.js
```diff
@@ -20,15 +20,28 @@ class ValidationError extends Error {
...
+
 export default {
   modules: {flights},
 
+  appConfigOverrides: {},
+  builderConfigOverrides: {},
+
+  configure(appConfig, builderConfig) {
+    configOverrides.appConfigOverrides = appConfig
+    configOverrides.builderConfigOverrides = builderConfig
```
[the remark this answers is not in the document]
jcbashdown: I'm not sure I understand how this would work. Moving it below the module registration would mean that the config overrides are not available when state()  gets called (which I think is during module registration right?)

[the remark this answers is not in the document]
jcbashdown: Maybe let's discuss in the catch up

src/main.ts
```diff
@@ -12,3 +15,9 @@ new Vue({
   store: new Vuex.Store({}),
   render: (h) => h(App),
 }).$mount('#app')
+
+// required inline due to front-loading of imports and having jQuery dependency
+import('jquery')
+  .then(() => {
+    import('bootstrap')
```
bzabos, line 22: Imports using this syntax are `async`/non-blocking. Do we want to `await` on these?

src/store/trees/bootstrap-legacy-global-dependencies.js
```diff
@@ -1,13 +1,14 @@
...
+import {merge, isEmpty} from 'lodash'
 
-export function bootstrapLegacyGlobalDependencies() {
+export function bootstrapLegacyGlobalDependencies(appConfig = null, builderConfig = null) {
   // initialize configuration sources
-  const __APP__ = require('../../../app.config')
-  const __CONTEXT__ = require('../../../builder.config')
+  const __APP__ = !isEmpty(appConfig) ? appConfig : require('../../../app.config')
+  const __CONTEXT__ = !isEmpty(builderConfig) ? builderConfig : require('../../../builder.config')
+  const source = appConfig.i18n ? appConfig.i18n : require('../../assets/messages.json')
```
bzabos, line 10: This logic changes slightly: `appConfig` may be `null` here. A couple possible options: either use lodash's `get(appConfig, 'i18n')` or default the param to an empty dictionary `{}` in the function signature.

src/main.ts
jcbashdown: I'm tried this but I was getting:

`Module parse failed: Cannot use keyword 'await' outside an async function`

Any tips? I've tried:

```
await import('jquery')
  .then(() => {
    await import('bootstrap')
  })
```

And 

```
await import('jquery')
  .then(() => {
    import('bootstrap')
  })
```

src/store/trees/bootstrap-legacy-global-dependencies.js
jcbashdown: Ah yes, meant to change it back to `{}` defaults

src/main.ts
jcbashdown: Discussed at Friday catch up - without await is fine for now.

bzabos (approved):
Approved, pending we sort out the `appConfig.i18n` potential for null in our configure action.

PR #17, opened 2020-11-05
VMO-2870 fix block-max-digit-editor

safydy-r:
a very quick fix:
- fix event on max-digit-editor
- fix some positive number filtering

### DIFF
--- a/src/components/common/NumericEditor.vue
+++ b/src/components/common/NumericEditor.vue
@@ -42,10 +42,15 @@
         type: [String, Number],
         required: true,
       },
+      regexNumericFiltering: {
+        type: String,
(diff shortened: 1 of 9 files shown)


src/components/common/NumericEditor.vue
```diff
@@ -42,10 +42,15 @@
         type: [String, Number],
         required: true,
       },
+      regexNumericFiltering: {
+        type: String,
+        required: false,
+        default: '[0-9\-]',
```
safydy-r, line 48: Numeric Editor should accept negative number by default

jcbashdown (approved):

PR #19, opened 2020-11-17
Vmo 3090 fix bugs found during qa round 1

safydy-r:
Will merge to master once approved


jcbashdown (approved):
Approved! However, Brett was going to add another checklist item to the ticket after our discussion on Tuesday, I think we can do this fix in another ticket though.

PR #20, opened 2020-11-19
Vmo 2627 delete outdated todos

jcbashdown:
Removing some todos here which I think are out of date or will naturally get tackled in tickets we already have. Let me know if you disagree with any of these.


bzabos:
Thanks for driving this @jcbashdown! I agree that having this work tracked in JIRA is a much more dependable way of ensuring we get to them eventually.

Can you do me one favour and link the tickets outputted from this task back to the original VMO-2627 so that we can refer back through history that way? (Thinking `relates to` or even `split from` could be good) Thanks!

bzabos (approved):

jcbashdown:
Will do, thanks Brett!

safydy-r:
Thanks working on these !
Can you pls comment why a TODO has been removed ? Which are `outdated` and which are `tackled in a specific ticket`, or which are simple questions that we already have clear answer?

jcbashdown:
Hey @safydy-r here is the reasoning for these:

`//TODO - is block editors the right place for this?`
 - My comment. I think now that the answer is yes
`//TODO - where should this come from - here and in other cases? Platform metadata? Need to write a story for this mode` 
- My comment again - we will handle this when we do the view mode. 
`<!-- TODO - Is passing around resouce the right way to do overrides? Is it even necessary? See comments in discoverContentTypesFor function -->` 
- No longer relevant as Brett changed the way we do this. 
`test: '', // todo: get started on a basic expression api`
- We'll handle this when we introduce the expression API - I feel like we aren't going to miss this.
`//TODO - waiting for Brett and Bart input on these prompts`
- This is a comment from me but I couldn't figure out what it meant. Maybe something about how these prompts are used during a call? Either way, they are in the spec and we've implemented them and we will work out how they are used when we do the flow runtime.

safydy-r:
Thanks @jcbashdown , seems good for me. Approving !

Issue #22, opened 2020-12-01
JQuery requirement error


smn:
Wanted to take this code base for a spin but am running into issues where after a successful build it seems to be missing a dependency on JQuery.

![Screenshot 2020-12-01 at 11 52 26](https://user-images.githubusercontent.com/1065/100731499-c451b980-33cb-11eb-8cbf-f26b7155f965.png)

Getting this in both Firefox and Chrome. Before I lose myself in Webpack and related build tools, is this a known issue?

safydy-r:
Thank you for pointing this. We're working on it. Will keep you informed

smn:
thanks! Appreciate that.

PR #23, opened 2020-12-09
VMO-2943 - seems to fix the jquery race condition

jcbashdown:
The first commit here is probably not ideal but it seems to fix the issue here:

https://github.com/FLOIP/flow-builder/issues/22

Let me know if you have another suggestion though Brett!

Probably more relevant for Safydy, in the second commit, I've taken the opportunity to fix stories which were broken in the package work. I've also moved them out of src - partly as I don't think they should be there and partly to prevent TS errors there breaking the dist build.

To avoid us breaking stories in future, I'm hoping this will keep us honest long term:

https://viamoinc.atlassian.net/browse/VMO-2631 

The last commit is a rebuild of dist. I don't think it's needed to fix that issue but we should do it for the new stuff we've merged to master recently. Once we're happy with this PR I will bump the version, publish and merge.


safydy-r:
@jcbashdown, any idea why the Toolbar stories was broken ? I tried to find but I have no idea so far
Screen Shot 2020-12-09 at 10 47 04 AM

jcbashdown:
@safydy-r I think that should now be fixed

safydy-r:
Thanks Jacob for the fix, approving

safydy-r (approved):

src/main.ts
```diff
@@ -3,23 +3,34 @@ import Vuex from 'vuex'
...
+async function initializeBootstrapFramework() {
+  import('bootstrap/dist/css/bootstrap.css')
+  import('bootstrap/dist/css/bootstrap-theme.css')
+
+  const {default: ImportedJquery} = await import('jquery')
+  exportAsGlobal({
+    $: ImportedJquery,
+    jQuery: ImportedJquery})
+
+  await import('bootstrap')
```
bzabos, line 23: Hi @jcbashdown, I did some testing using DevTools' throttling and was still able to reproduce. I've opted to push these guts into an async function to handle the dependencies more explicitly.

src/main.ts
```diff
@@ -3,23 +3,34 @@ import Vuex from 'vuex'
...
-    .then((jQuery) => {
-      import('bootstrap')
-    })
+function exportAsGlobal(exports) {
+  Object.assign(window, exports)
+}
+
+async function initializeBootstrapFramework() {
+  import('bootstrap/dist/css/bootstrap.css')
+  import('bootstrap/dist/css/bootstrap-theme.css')
```
bzabos, line 16: I've pulled these imports down into this function as well so our imminent bootstrap-4.5 migration has everything laid out.

bzabos (approved):
I didn't push up another build commit, because I didn't know what you'd meant by build 😹 . My assumption was that for now "build" implies `build-bundle` producing a lib consumable, is that aligned with your thoughts as well?

Approving. Let me know whether or not you'd like to hash out my changes before merging.

smn:
I'm able to run this now! I am getting a build error (which may be due to me having missed some steps) but the UI is loading. Thank you! 

Screenshot 2020-12-11 at 10 31 42

jcbashdown:
Thanks @bzabos! That looks much better. Thanks for pointing that out also @smn . I'm also getting that but didn't notice it before. Investigating now.

jcbashdown:
@smn Looks like we needed to add a loader declaration for css. Not sure why it wasn't necessary before. Please try again now.

@bzabos - you're right - build-bundle is the first step. Then we need to bump the version in package.json and publish to npm for use in voto5. When we're happy with this I'll go ahead with that. We still have the ticket outstanding to come up with our proper release process but this is good enough for now for the integration work we're focusing on.

smn:
@jcbashdown works for me now! Thank you!

jcbashdown:
Great, thanks for your patience on this Simon. Bumped the version and published. Merging now.

PR #26, opened 2021-02-02
VMO-3432 fix-type-c-does-not-satisfy-the-constraint

safydy-r:
It was a ts issue on storybook. I'm wondering why we didn't have it before?
My proposal is just to exclude storybook from `tsconfig.json`


jcbashdown (approved):
Does the trick for me

PR #28, opened 2021-02-11
Vmo 3371 drop added blocks relative to the active block

safydy-r:


src/store/builder/index.ts
```diff
@@ -354,3 +354,22 @@ export function generateConnectionLayoutKeyFor(source: IBlock, target: IBlock) {
...
+    xPosition = get(block, 'platform_metadata.io_viamo.uiData.xPosition', 0) + xDelta
+    yPosition = get(block, 'platform_metadata.io_viamo.uiData.yPosition', 0) + yDelta
+  } else {
+    // put in the viewport center
+    let builderCanvasElement = document.getElementsByClassName('builder-canvas')[0]
+    let sideBarElement = document.getElementsByClassName('tree-sidebar-container')[0]
+    const rect = builderCanvasElement.getBoundingClientRect()
+
+    xPosition = Math.round(Math.abs(rect.left) + (window.innerWidth - sideBarElement.clientWidth) / 2)
+    yPosition = Math.round(Math.abs(rect.top) + window.innerHeight / 2)
```
bzabos, line 371: Hi @safydy-r, we're sourcing `xPosition` and `yPosition` from three places, and still end up with a default of {0,0} when positional data absent from block :) Is there a way we can do this in two phases to help with readability and reduce branching?

1. discover {x,y} from block
2. default x to viewport center when absent or empty
3. default y to viewport center when absent or empty
4. return {x,y} + delta

---

Pseudo eg. 
```
func computeBlockPositionsFrom(block)
  x = get(block, uiData.xPosition)
  y = get(block, uiData.yPosition)

  when x is empty:
    x = viewportCenter.x

  when y is empty:
    y = viewportCenter.y

  return {x + xDelta, y + yDelta}
```

src/store/builder/index.ts
```diff
@@ -354,3 +354,22 @@ export function generateConnectionLayoutKeyFor(source: IBlock, target: IBlock) {
     ...map(target.exits, 'tag'),
   ]
 }
+
+export function computeBlockPositionsFrom(block: IBlock) {
```
bzabos, line 358: Does this break when null/empty is provided? 
Do we need the type to specific null/undefined?

bzabos (approved):
This is perfect @safydy-r ! I left a comment to see if we could simplify a bit, but I appreciate your thoroughness in the case when a block has not been selected. Nice!

src/store/builder/index.ts
safydy-r: no it won't break.
The only scenario we will get `null` is when there is no activated block. Changed to `IBlock | null`

adeablay (approved):

PR #30, opened 2021-02-23
Vmo 3368 dynamic canvas size

safydy-r:


src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,6 +26,50 @@
 
     computed: {
       ...mapState('flow', ['flows']),
+
+      canvasHeight() {
+        const defaultHeight = 1500
```
poojahpatil90, line 31: Would it be better to use `screen.availHeight` instead of hardcoding the height to a number?

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,6 +26,50 @@
...
+        if (!blockElement) {
+          return defaultHeight
+        }
+
+        const blockHeight = blockElement.clientHeight
+        return lodash.get(blockAtTheLowestPosition, 'platform_metadata.io_viamo.uiData.yPosition') + blockHeight + 100
+      },
+
+      canvasWidth() {
+        const defaultWidth = 1500
```
poojahpatil90, line 53: Same as above maybe we could use `screen.availWidth`?

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: This is a great proposition @poojahpatil90 , I'm making a change

poojahpatil90 (approved):

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,10 +26,69 @@
...
+
+        const blockWidth = blockElement.clientWidth
+        return lodash.get(blockAtTheFurthestRightPosition, 'platform_metadata.io_viamo.uiData.xPosition') + blockWidth + 100
+      },
     },
 
     mounted() {
       this.$store.dispatch('builder/loadFlow')
+
+      // Auto scroll, needed to put in mounted()
```
safydy-r, line 78: @bzabos , I thought about the auto scroll you mentioned during the demo and I found an idea to implement it

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,10 +26,69 @@
...
     computed: {
       ...mapState('flow', ['flows']),
+
+      canvasHeight() {
+        const defaultHeight = window.screen.availHeight
+        if (!this.flows[0].blocks && this.flows[0].blocks.length) {
+          return defaultHeight
+        }
+
+        const blockAtTheLowestPosition = lodash.maxBy(this.flows[0].blocks, 'platform_metadata.io_viamo.uiData.yPosition')
```
bzabos, line 36: I never realized how backwards this variable is before! `yPosition` is inverted if we're thinking of a cartesian plane, right? Where an upward direction is our position y-value and downward would be a negative or `min()` y-value! 

Did you find that confusing in this task?

bzabos: Ps. Can we swap to `import {maxBy}` from 'lodash' rather than `lodash.maxBy()`?

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,10 +26,69 @@
...
+
+        if (!blockAtTheLowestPosition) {
+          return defaultHeight
+        }
+
+        const blockElement = document.getElementById(`block/${blockAtTheLowestPosition.uuid}`)
+
+        if (!blockElement) {
+          return defaultHeight
```
bzabos, line 45: Can we throw a `console.log()` onto this branch? It feels like this would be an odd state to get into! 
Something like:`log('Interaction Designer', 'Unable to find dom element corresponding to lowest block id.')`

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,10 +26,69 @@
...
+        }
+
+        const blockHeight = blockElement.clientHeight
+        return lodash.get(blockAtTheLowestPosition, 'platform_metadata.io_viamo.uiData.yPosition') + blockHeight + 100
```
bzabos, line 49: Can we `import {get} from 'lodash'` here?

bzabos: Can we assign this `100` to a variable so that we have some indication to the fact that it's a buffer beyond our furthest outlying block?

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -44,9 +106,4 @@
     -ms-user-select: none; /* Internet Explorer/Edge */
     user-select: none; /* Non-prefixed version, currently supported by Chrome, Opera and Firefox */
   }
-
-  .builder-canvas {
-    width: 9999px;
-    height: 9999px;
```
bzabos, line 50: Do we want to leave a standard proportion min-width and min-height here so that initial render on slower machines take up space? (something like 1024x768 could be fine?)

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,10 +26,69 @@
...
+        window.scrollTo({
+          top: newValue,
+          behavior: 'smooth'
+        })
+      }, { immediate: true });
+
+      this.$watch('canvasWidth', (newValue) => {
+        window.scrollTo({
+          left: newValue,
+          behavior: 'smooth'
```
bzabos, line 89: Nice one @safydy! This is such a nice + simple solution! 

Can you help me understand a handful of things?
1. Does this then scroll the entire page over by the entire width (for eg) as soon as I touch the edge? Would you be willing to attach a gif or vid of this in action on the PR description? Thanks!
2. Can we debounce/throttle the `scrollTo()` invocations so they're not called too rapidly?
3. Can you help me understand what influenced the decision to set up watchers in `mounted()` rather than as a top-level definition as per: https://vuejs.org/v2/guide/computed.html#Watchers ?
4. Is `{ immediate: true }` absolutely necessary here?

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -23,10 +26,69 @@
...
+        }
+
+        const blockElement = document.getElementById(`block/${blockAtTheFurthestRightPosition.uuid}`)
+
+        if (!blockElement) {
+          return defaultWidth
+        }
+
+        const blockWidth = blockElement.clientWidth
+        return lodash.get(blockAtTheFurthestRightPosition, 'platform_metadata.io_viamo.uiData.xPosition') + blockWidth + 100
```
bzabos, line 71: I think it could be important for us to separate these methods into a few parts to allow vuejs to more efficiently trigger recache of these as computed properties. A couple things I'm thinking about: 
1. Are we mixing domain data lookup and UI manipulation (eg. dom access, window dimensions, etc)
2. How frequently is the underlying data that we're accessing going to change? 

A suggestion:
Create separate computer getters for things returned from dom access and those that specifically do analysis over the underlying data (eg. maxBy):
- `largestXPositionedBlockId()`
- `largestXPositionedBlock()` — Set up refs on block and index into that locally via `$refs[largestXPositionedBlockId()]`; this could prevent having to call `getElementById()` at all. 
- `windowWidth()` — I'm actually wholly uncertain as to how vuejs can trigger on changes to window properties, keep an eye out for this as a place where changes and redraws may not be propagating as anticipated. An alternative could be to observe window's resize event and set that explicitly so that we've got that cached locally.

Any accesses to the DOM should be considered equivalent to us going to the DB on the server side, fast, but very slow compared to having things local or cached. They're often fast enough, but in situations like this where we're dispatching many events per second while a person is dragging, and doing work to repaint connections, we need to pay attention to efficiency. 

Let me know if you'd like to chat this one out! 😸

bzabos (changes requested):
Thanks @safydy! This looks great! 

I have made a couple suggestions that are not functional, but primarily separation of concerns.

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: sure, added it

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: Sure, added it as `deltaY`

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: about cartesian plane, I thought about a possible inversion, but I ended with no needed changes. In fact,
In this particular code, we're trying to get Width & Height of the canvas. Meaning, we just need to translate the `yPosition` into a specific Height, and `xPosition` into Width

This is what we get when adding a block:
```
from `platform_metadata.io_viamo.uiData`
{
"xPosition":635,
"yPosition":329
}

related `blockElement`:
clientWidth: 635
clientHeight: 329
```

Therefore, the maxBy is still accurate with positive sign as we look for an absolute value.

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: this is a good idea, I added a default value here for min-width, min-height

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: @bzabos : 
Thanks 😊 ,
- for 1 & 2: I have thought about adding debounce when implementing, but it seems not working correctly. Here are 03 videos about 03 solutions : scroll smooth, scroll auto (not smooth), scroll auto with debounce (tried many delays) [https://drive.google.com/drive/folders/10oyAphd1mLK60kthVqQoLH6p2ji5IQdv?usp=sharing](https://drive.google.com/drive/folders/10oyAphd1mLK60kthVqQoLH6p2ji5IQdv?usp=sharing) 

> I opted for the **scroll smooth** solution

- for 3: outside mounted, I had an unavailable prop. But just check it and found I made a mistake on certain condition to return default value. Not it's working, and we don't need to put it inside mounted. Thanks for insisting here.
- for 4. I think it's not necessary, removed it.

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: @bzabos , I think I need a catchup about `$refs[largestXPositionedBlockId()]` & `windowWidth()`

src/components/interaction-designer/BuilderCanvas.vue
bzabos: @bzabos is going to leave an example of how we could use `debounce()` to support this feature.

src/components/interaction-designer/BuilderCanvas.vue
bzabos: Hi @safydy, here are the resources we'd discussed: https://developers.google.com/web/fundamentals/performance/rendering

src/components/interaction-designer/BuilderCanvas.vue
bzabos: To debounce a method: we need to cache a reference to the debounced version, then invoke that debounced version. I think this example could help: 
```
export default {
  methods: {
    debounceMethod: _.debounce(() => {
      console.log('Debounce button clicked!')
    }, 2000)
  }
}
```

via: https://www.digitalocean.com/community/tutorials/vuejs-lodash-throttle-debounce#debouncing-methods

safydy-r: I used debounce for now to give us more control on the `timer`

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: done

safydy-r:
@bzabos , do you need more time to think about this, or you're happy we move it to QA?

bzabos (approved):
As per our call; this is great! Thanks @safydy!

PR #31, opened 2021-02-23
VMO-3369 toggle view and edit

poojahpatil90:
VMO-3369

Builder canvas - Read-only / view-mode of designer canvas elements (+ hook up Edit Flow button without page reload)


src/components/common/AudioLibrarySearchField.vue
```diff
@@ -69,102 +70,102 @@
...
-      },
-
-      hasPrevious() {
-        return this.offset > 0
-      },
-		},
-
-		methods: {
-			search(query) {
+import fuse from 'fuse.js';
```
safydy-r, line 73: @bzabos , can you remind us our rule for semi-colon ?
If we expect trailing semi, shall we add a rule in `eslintrc` to avoid this kind of change ?

src/components/common/AudioLibrarySearchField.vue
```diff
@@ -69,102 +70,102 @@
 </template>
 
 <script>
-	import fuse from 'fuse.js'
```
safydy-r, line 72: @poojahpatil90 , thanks for thinking of these indentation etc, but I would avoid making auto fix like this before PR review. This is a big distraction for reviewer especially for big PR.

My suggestion is to wait for PR approval, then commit on last change to make auto fix from IDE before merge, what do you think?

src/components/common/AudioLibrarySearchField.vue
```diff
@@ -3,6 +3,7 @@
     <div class="input-group">
       <span class="input-group-prepend">
         <button :class="{active: isEntireLibraryModeEnabled}"
+                :disabled="!isEditable"
```
safydy-r, line 6: This is good for a 1st PR in flow-builder.
I noticed you introduced `isEditdable` prop in many components to disable the side bar children.
I would suggest we take advantage of the `<fieldset>` like here:  https://bitbucket.org/voto/voto5/src/e642ee81827f72c654cced81c9e304e80e4d64ea/resources/assets/js/trees/components/block-types/ValidateCodeBlock.vue#lines-10

This way the maintenance will be less painful because we don't need to add `isEditable` for future block config elements. 

Here is a overview how to disable it
- change the element having `tree-sidebar` class into `<fieldset>`
- put conditional disabled there
Screen Shot 2021-02-23 at 1 09 39 PM

safydy-r: But pls get confirmation from @bzabos first if he is good with this suggestion

src/components/common/AudioLibrarySearchField.vue
poojahpatil90: @safydy-r  Sorry about the indentation fixes. My IDE was setup to auto fix the ESLint issues, which I did not realize until a later point and it was difficult to revert back those changes. 
One solution for now is to hide white-space changes from the settings button above(as suggested by Brett)

src/components/common/AudioLibrarySearchField.vue
poojahpatil90: @safydy-r  Thanks for your suggestions. I will try out fieldset and update the PR.

poojahpatil90:
I have removed the semicolon changes on this PR. Please choose 'hide whitespace changes' available on the settings button.  Although there are some other changes that I could not prevent, it will be much better to go over my changes now.

stories/Toolbar.stories.ts
```diff
@@ -43,17 +44,18 @@ class BaseMountedClass extends Vue {
...
+  @builderVuexNamespace.Action setIsEditable: void
 }
 
 // Default
 @Component<any>(
     {
         ...BaseOptions,
       async mounted() {
-        this.updateIsEditable({value: 0})
+        this.setIsEditable(0)
```
safydy-r, line 58: @poojahpatil90 , let's pass boolean type as defined in IBuilderState

src/store/builder/index.ts
```diff
@@ -15,86 +18,86 @@ export enum OperationKind { // todo migrate these to flight-monitor
...
+    [OperationKind.CONNECTION_SOURCE_RELOCATE]: IConnectionSourceRelocateOperation;
+    [OperationKind.CONNECTION_CREATE]: IConnectionCreateOperation;
+    [OperationKind.BLOCK_RELOCATE]: null;
+  };
 }
 
 export const stateFactory = (): IBuilderState => ({
   activeBlockId: null,
-
+  isEditable: null,
```
safydy-r, line 57: @poojahpatil90 , let's put `true` by default. Otherwise all stories for blocks in storybook will have **disabled** block config

tests/unit/__snapshots__/storybook.spec.ts.snap
```diff
@@ -143,15 +143,17 @@ exports[`Storyshots InteractionDesigner/Toolbar Default 1`] = `
                        
                       <!---->
```
safydy-r, line 145: @poojahpatil90 , I'm happy you quickly learnt how to handle this, good job

stories/Toolbar.stories.ts
poojahpatil90: Yes, I missed that. I will make the change.

src/store/builder/index.ts
poojahpatil90: Sure, I will make this change.

tests/unit/__snapshots__/storybook.spec.ts.snap
poojahpatil90: Thanks!  My knowledge on storybook is still limited. I am looking forward to learning and playing around with it more.

safydy-r (approved):

src/components/common/AudioLibrarySearchField.vue
bzabos: Good one, I've been pushing for us to skip semi-colons!

bzabos (approved):
Thanks @safydy for the recommendation to use fieldsets! That made me smile :) To note: We'll need to explicitly pass `isEditable` for some non-native html components, but overall I favour this approach as well.

Looks good @poojahpatil90, thanks!

src/components/interaction-designer/flow-editors/ExitBlockEditorToggle.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="flow-editor exit-block-editor-toggle">
-    <template v-if="true || isEditable && hasClipboard">
+    <template v-if="isEditable && hasClipboard">
```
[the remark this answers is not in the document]
poojahpatil90: Added a note to handle this part as part of VMO-3546(for improving the view mode)

src/components/common/PlainDraggable.vue
```diff
@@ -2,97 +2,95 @@
...
+    const handle = this.$el.querySelectorAll('.draggable-handle')[0]
+
+    this.draggable = new PlainDraggable(this.$el, {
+      containment: document.querySelector('.builder-canvas'),
+      autoScroll: true,
+
+      // prevent css translate() animations for move
+      // they don't seem to be throttled enough for leaderline to follow tightly
+      leftTop: false,
+      disabled: true,
```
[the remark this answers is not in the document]
poojahpatil90: We could standardize this behavior if it makes sense :)

src/store/builder/index.ts
```diff
@@ -15,86 +18,86 @@ export enum OperationKind { // todo migrate these to flight-monitor
...
+    [OperationKind.CONNECTION_SOURCE_RELOCATE]: IConnectionSourceRelocateOperation;
+    [OperationKind.CONNECTION_CREATE]: IConnectionCreateOperation;
+    [OperationKind.BLOCK_RELOCATE]: null;
+  };
 }
 
 export const stateFactory = (): IBuilderState => ({
   activeBlockId: null,
-
+  isEditable: true,
```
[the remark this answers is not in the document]
poojahpatil90: Done!

src/views/InteractionDesigner.vue
```diff
@@ -44,201 +44,205 @@
...
+    console.debug('Vuej tree interaction designer mounted!')
+  },
 
-          invoke(exported, 'install', this)
-          Vue.component(`Flow${typeWithoutSeparators}`, exported.default)
-        })
-      },
+  watch: {
+    mode(newMode) {
+      this.updateIsEditableFromParams(newMode)
```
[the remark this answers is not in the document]
poojahpatil90: You are right, it is not ideal to use watchers. I do not like using them either, but in this case nothing else worked for me(at the time I was coding). I will take some time to explore other ways to catch changes on props effectively. I have added a note to tackle this as part of VMO-3546.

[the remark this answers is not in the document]
bzabos: Thanks! And you can always reach out to me if you want to brainstorm ways to avoid watchers :)

PR #35, opened 2021-03-03
Vmo 3120 flow lifecycle

jcbashdown:
There a number of TODOs added here which seem like open questions for discussion or things for future tickets to me. Would be good to get your POV.

Embedding in Voto5 is covered here: https://bitbucket.org/voto/voto5/pull-requests/1787/vmo-3120-flow-lifecycle


src/views/Home.vue
```diff
@@ -0,0 +1,57 @@
...
+@Component<any>(
+  {
+    mixins: [lang, Routes],
+    async created() {
+      const {$store} = this
+
+      forEach(store.modules, (v, k) =>
+        !$store.hasModule(k) && $store.registerModule(k, v))
+
+      this.configure({})
```
bzabos, line 44: It seems like this piece created the conditions for needing `isConfigured`— Is there a reason we're preferring not to do a full configure here? // How come this `created()` is differs slightly from the others?

src/views/InteractionDesigner.vue
```diff
@@ -193,23 +195,31 @@ export default {
 
   /** @note - mixin's mount() is called _before_ local mount() (eg. InteractionDesigner.legacy::mount() is 1st) */
   mounted() {
+      this.flow_setActiveFlowId({flowId: this.id})
+
+      //if nothing was found for the flow Id
+      if(!this.activeFlow) {
+        this.flow_setActiveFlowId({flowId: null})
+        this.$router.replace(`${this.route('flows.fetchFlow', {flowId: this.id})}?nextUrl=${this.$route.path}`)
```
bzabos, line 203: I believe we can get the router to `querystring` serialize our query params for us. Something like this should ensure safely urlencoded param value:
```
const {id: flowId} = this
const {path: nextUrl} = this.$route

this.$router.replace(
    this.route('flows.fetchFlow', {flowId}), 
    {query: {nextUrl}})
```

src/views/NewFlow.vue
```diff
@@ -0,0 +1,96 @@
...
+            <div v-if="flowError" class="alert alert-danger" role="alert">
+              {{flowError | trans}}
+            </div>
+            <flow-editor :flow="activeFlow" flow-header="flow-builder.create-flow" :sidebar=false />
+
+            <div class="float-right">
+              <router-link :to="route('trees.editTree', {treeId: activeFlow.uuid, component: 'interaction-designer', mode: 'edit'})"
+                class="btn btn-primary"
+                event=""
+                @click.native.prevent="handlePersistFlow(route('trees.editTree', {treeId: activeFlow.uuid, component: 'interaction-designer', mode: 'edit'}))">
```
bzabos, line 16: Does this need to be a router-link? It seems like we prevent the router-link from doing its natural thing and do our own routing below. Should this just be a vanilla button?

src/store/flow/flow.ts
```diff
@@ -16,20 +17,69 @@ import moment from 'moment'
...
+      uuid: "TODO",
+      name: "TODO",
+      description: "TODO",
+      platform_metadata: {},
+      flows: state.flows,
+      resources: state.resources
+    } as unknown as IContext
+  },
+  flowsList: state => {
+    return state.flows
```
bzabos, line 57: Out of curiosity: Would you prefer to keep this getter rather than removing + doing a `mapState` directly in the component that consumes `flowsList` ? 

The reason I'm asking is that it feels like a tiny bit of added indirection in a space where we haven't reconciled `IContext`, `IContainer`, `activeFlowContainer` (above) — Might be nice to mitigate having yet another way to find a list of flows 🤷‍♀️ 😸

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -30,14 +30,15 @@
...
-                         class="btn btn-outline-secondary mr-2"
-                         :title="trans('flow-builder.click-to-toggle-editing')"
-            >
+          <router-link v-if="!ui.isEditableLocked"
+             :to="editOrViewTreeJsUrl"
+             event=""
+             :title="trans('flow-builder.click-to-toggle-editing')"
+             class="btn btn-outline-secondary mr-2"
+             :class="{active: isEditable}"
+             @click.native.prevent="handlePersistFlow(editOrViewTreeJsUrl)">
```
bzabos, line 39: Does this end up calling save even when toggling from View Mode into Edit Mode ?

stories/CaseBlockStyled.stories.ts
```diff
@@ -22,7 +22,7 @@ export default {
 }
 
 const CaseBlockTemplate = `
-  <flow-builder-sidebar-editor-container :block="activeBlock">
+  <flow-builder-sidebar-editor-container v-if="activeFlow" :block="activeBlock">
```
bzabos, line 25: I think we've already got something like this partially covered from inside of `FlowBuilderSidebarEditorContainer.vue`, like:
```
<flow-builder-container v-if="block || flow">
```

Are we duplicating the logic? Could be nice to not have to remember to add this `v-if` on new block types.

src/store/trees/trees.js
```diff
@@ -50,12 +51,14 @@ export default {
...
         ],
+        saveCurrentlyInProgress: 0,
+        isEditable: true
       },
     }
   },
 
   getters: {
-
+    isConfigured: ({ui}) => ui.isConfigured,
```
bzabos, line 61: This is declared above as `configured: false` — Should this say `ui.configured` or the others be `isConfigured`?

src/store/trees/trees.js
```diff
@@ -50,12 +51,14 @@ export default {
         enabledFeatures: [
           /** @see \Voto5\Http\Controllers\V3TreesController::get_editTree */
         ],
+        saveCurrentlyInProgress: 0,
+        isEditable: true
```
bzabos, line 55: Can we rename our `created` property to `isCreated` / `is_created` to indicate boolean? I'd initially read that as a date/timestamp, equivalent to something like `createdAt`. 

`saveCurrentlyInProgress` → `isSaveCurrentlyInProgress`
`configured` → `isConfigured`

bzabos (changes requested):
Hi @jcbashdown, this looks great! I've left a pile of general/non-urgent/context comments, along with 5 suggestions that would be nice to get in: 
- [x] use async/await rather than promises
- [x] use null for absent active flow
- [ ] use bool naming convention + bool value for newly added flags (rather than number)
- [ ] update `activeFlowContainer` to match IContainer
- [ ] log errors into the console in `catch()` blocks

But overall: wow! A big piece we've been talking about for a while, I'm excited to see this settle in as part of the builder's experience 👌

src/store/flow/flow.ts
```diff
@@ -16,20 +17,69 @@ import moment from 'moment'
...
 export const getters: GetterTree<IFlowsState, IRootState> = {
-  activeFlow: (state) => state.flows.length && getActiveFlowFrom(state as unknown as IContext),
-
+  //We allow for an attempt to get a flow which doesn't yet exist in the state - e.g. the firstFlowId doesn't correspond to a flow
+  activeFlow: (state) => {
+    if(state.flows.length) { 
+      try {
+        return getActiveFlowFrom(state as unknown as IContext)
+      } catch(err) {
+        return 0
```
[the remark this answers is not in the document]
jcbashdown: This was me trying to preserve the original type I was finding on master to avoid any weirdness but let me just try and switch to a bool and see how it goes. I think QA will pick up any issues I miss.

src/store/flow/flow.ts
jcbashdown: Hmm. I think I get what you are saying. The Getter isn't necessary yet so let's move to mapState.

src/store/flow/flow.ts
```diff
@@ -16,20 +17,69 @@ import moment from 'moment'
...
+      }
+    }
+  },
+  //TODO - is the IContext equivalent to the Flow Container? Can we say that it should be?
```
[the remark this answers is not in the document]
jcbashdown: In your checkbox you have "update activeFlowContainer to match IContainer" but what I'm taking from this is that this isn't something we can do quite yet..? It also might not be easy in this PR if we want to avoid conflicts with Safydy's flow-runner upgrade PR perhaps?

src/store/flow/flow.ts
```diff
@@ -93,8 +143,8 @@ export const mutations: MutationTree<IFlowsState> = {
     const block: IBlock = findBlockWith(blockId, flow) // @throws ValidationException when block absent
     Vue.set(flow, 'firstBlockId', block.uuid)
   },
-  flow_setName(state, { flowId, value }) {
-    findFlowWith(flowId, state as unknown as IContext).name = value
+  flow_setNameFromLabel(state, {flowId, value}) {
+    findFlowWith(flowId, state as unknown as IContext).name = value.replace(/\W+/g, '')
```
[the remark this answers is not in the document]
jcbashdown: I'll switch to label. For the replace, this is taken from what we have for block names (we said we would follow the rules for this in the ticket)

> Update the src/components/interaction-designer/block-editors/LabelEditor.vue component to update the flow name behind the scenes when the label is edited to a string conforming to the same rules as block-editors/NameEditor.

src/store/flow/flow.ts
```diff
@@ -117,6 +167,43 @@ export const mutations: MutationTree<IFlowsState> = {
 }
 
 export const actions: ActionTree<IFlowsState, IRootState> = {
+
+  async flow_persist({ state, getters, commit }, { persistRoute, flowContainer }): Promise<IContext> {
+    const restVerb = flowContainer.created ? 'put' : 'post'
+    const oldCreatedState = flowContainer.created
+    return axios[restVerb](persistRoute, omit(flowContainer, ['created']))
+      .then(({data}) => {
```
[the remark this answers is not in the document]
jcbashdown: If it's okay with you, I'd like to leave this for now. Not super familiar with flight monitor and I want to unblock other PRs!

Here's a place holder for this:

https://viamoinc.atlassian.net/browse/VMO-3768

src/store/flow/flow.ts
```diff
@@ -117,6 +167,43 @@ export const mutations: MutationTree<IFlowsState> = {
...
+        commit('flow_updateFlowContainer', data)
+        commit('flow_updateCreatedState', true)
+        return getters.activeFlow
+      })
+      .catch((error) => {
+        if(!fetchRoute) {
+          console.info("Flow fetch route not configured correctly in builder.config.json. Falling back to vuex store")
+          return getters.activeFlow
+        }
+        return null
```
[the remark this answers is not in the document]
jcbashdown: Done

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
jcbashdown: Good point. I think it does. Fixing this.

stories/CaseBlockStyled.stories.ts
jcbashdown: For some reason this wasn't quite working for me. Let me try again...

src/store/trees/trees.js
jcbashdown: Good spot. I actually just found a bug where configure was incorrectly being called multiple times. This fixes that!

src/store/trees/trees.js
jcbashdown: Done

stories/CaseBlockStyled.stories.ts
jcbashdown: Okay, pushed a fix for this:

https://github.com/FLOIP/flow-builder/pull/35/files#diff-130b7357fda7a41c13a1ec6be92137441c79e5a81a4c8ca6445ad1c1cd75ae9fR75

This crashes if no flow is yet present. Slightly unclear why this wasn't an issue before...

jcbashdown:
@bzabos I believe I've not addressed all your comments. I've found one small flow switching error while doing this which I'm looking at now. Other than that I'm hoping this is enough to move to QA. Please let me know if I missed or misunderstood something though!

src/store/flow/flow.ts
```diff
@@ -16,20 +17,63 @@ import moment from 'moment'
...
+    return {
+      isCreated: state.isCreated,
+      specification_version: "TODO",
+      uuid: "TODO",
+      name: "TODO",
+      description: "TODO",
+      platform_metadata: {},
+      flows: state.flows,
+      resources: state.resources
+    } as unknown as IContext
```
jcbashdown, line 54: This will become IContainer when the flow-runner version bump is merged. These TODO fields all need to be deleted for now based on discussion in backend runtime hilda. CC @bzabos . I'll add in the latest spec version now it's settled a bit. Is there anywhere good we can pull this from dynamically?

jcbashdown: - Will do this in when moving to IContainer after this is merged to master and the flow-runner version bump is also merged there.

src/views/NewFlow.vue
```diff
@@ -0,0 +1,95 @@
...
+import {IFlow, IContext} from '@floip/flow-runner'
+
+@Component<any>(
+  {
+    components: {
+      FlowEditor,
+    },
+    mixins: [lang, Routes],
+    async mounted() {
+        await this.flow_addBlankFlow()
```
jcbashdown, line 47: @bzabos Found a bug with this today that I've been wrestling with but I think I've not decided to leave fixing for now. Basically, when you hit this page a blank flow is added to the container. When you "Save and continue" we hit the backend dummy route for persisting and move to the interaction designer. The issue is, if you came to this page from the link on Home.vue and then instead of submitting you go back to the homepage with the back button (which vue-router handles) you will see the empty flow in the list there - even though you never saved it.

I think a fix for this is a bit tricky as all of the flow-editing components edit the currently active flow which comes from the active flow container. It would be a big-ish refactor to let these components edit this or alternatively edit a temporary flow that is not yet in the container. This might also result in a pretty ugly component. 

The alternative is to set some sort of isPersisted/isCreated flag on the flows when we persist and then not render un persisted/created flows on the home page. Then we can find some point in the flow lifecycle to clear out unpersisted flows. This feels a bit hacky though.

As I understand it, we may in future have a draft state that we can use here and do something like option 1 in a more though out way. This coupled with the fact that this Home component won't actually ever be used by voto5 (the flows index will end up coming from php there and any navigation to this page will pull only the actually persisted flows) means I don't think this is a big enough issue to invest time in fixing now. Let me know if you disagree though.

src/views/FetchFlow.vue
```diff
@@ -0,0 +1,83 @@
...
+import lang from '@/lib/filters/lang'
+import Routes from '@/lib/mixins/Routes'
+import { Component, Prop } from 'vue-property-decorator'
+import Vue from 'vue'
+import {Getter, Mutation, namespace} from 'vuex-class'
+import {store} from '@/store'
+const flowVuexNamespace = namespace('flow')
+import {IFlow} from '@floip/flow-runner'
+
+@Component<any>(
```
safydy, line 31: @jcbashdown , can you try removing this `<any>` for new added components, I think it won't be necessary. (For existing component, I removed those `<any>` on vmo-2623 PR https://github.com/FLOIP/flow-builder/pull/55/files)

src/store/flow/flow.ts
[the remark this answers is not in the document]
bzabos: Ohh, I see! From my perspective, it would be really nice to see "Hello quick brown fox" result in "hello_quick_brown_fox", I don't know what ticket that would be a part of, though. If it doesn't feel like it should be a part fo this changeset, that's also sounds totally fine to me.

src/store/flow/flow.ts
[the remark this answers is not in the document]
bzabos: Sounds great!

src/views/FetchFlow.vue
bzabos: CC/ @poojahpatil90

src/views/NewFlow.vue
bzabos: Nice! To your conclusion: I agree with you, no need to fix for now! Can we create a JIRA to revisit this in `2.0`-ish?

Overall, I'm partial to the `draft` concept where we would edit a local flow, then hitting _Continue_ would then add the flow to the state. 

Begs the question of whether or not we should update our approach to vuex actions API to be more of a functional toolset like lodash, eg: `setFlowName("My First Flow", {flow})` which would then modify the _reference_ (rather than performing lookup from state) while simultaneously continuing to provide an encapsulation of any logic needed behind the scenes.

stories/CaseBlockStyled.stories.ts
bzabos: This link doesn't end up landing me on anything for me 😛 !

bzabos: I think my question is whether we should update the guard that already exists and would only need a change in one place, rather than adding an additional guard to every story?

src/store/flow/flow.ts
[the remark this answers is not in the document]
bzabos: My suggestion was to update the set of properties listed below to match the properties on flow runner's `IContainer` to pave way for when we do have access to `IContainer`.

If you're suggesting that this affects more than just a couple places, then yes let's leave it for now!

bzabos (approved):

bzabos:
@jcbashdown I've switch my review to Approved and left a couple last comments. I'm leaving it up to your discretion as to how to handle those. Thanks for taking the time to respond to the feedback!

src/store/flow/flow.ts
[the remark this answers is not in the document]
jcbashdown: Interesting. I was basing this on the spec. I'll take a quick look at the differences from IContainer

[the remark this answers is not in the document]
jcbashdown: Hmm. I can't find IContainer. Where is this..? Not seeing it in the latest flow-runner.

stories/CaseBlockStyled.stories.ts
jcbashdown: Yup, exactly. Not sure why that link doesn't load. Let me try again:

https://github.com/FLOIP/flow-builder/pull/35/files#diff-130b7357fda7a41c13a1ec6be92137441c79e5a81a4c8ca6445ad1c1cd75ae9fR85

That link works for me to show what I needed to make the existing guard work. This allowed me to delete all changes to stories I had made previously which you can now see in the diff (which no longer includes any stories)

src/views/Home.vue
jcbashdown: This is just a shortcut really. To be honest it's probably a bit lazy. We decided that this component is only used in the builder and intended to be overridden when using the app. I actually think that won't always be the case. I'm going to add in a full configure now.

PR #37, opened 2021-03-08
Vmo 3320 allow block types imported from outside

safydy-r:
To test the current PR, I created a branch in voto5, check this PR for more details https://bitbucket.org/voto/voto5/pull-requests/1852/vmo-3320-allow-block-types-imported-from


builder.config.json
```diff
@@ -37,91 +37,91 @@
       }
     ],
     "blockClasses": {
-      "ConsoleIO\\Print": {
+      "components/interaction-designer/block-types/ConsoleIO_PrintBlock": {
```
safydy-r, line 40: @bzabos , wepback seems to require a minimun static path prefix to accept dynamic variable (eg "`@/`"), https://github.com/FLOIP/flow-builder/pull/37/files#diff-4778731e3debd774a6c14025179b9bfb34aae3d45fb42448205377ca6475503dR219

So I didn't add `@` here, let me know if this works as you expected

src/store/trees/trees.js
```diff
@@ -79,6 +79,8 @@ export default {
 
 		interactiveBlockClasses: ({ui}, getters, rootState) => lodash.pickBy(ui.blockClasses, (value, key) => value.is_interactive),
 
+    blockClassFromBlockType: ({ui}, getters, rootState) => (blockType) => lodash.values(lodash.pickBy(ui.blockClasses, (value, key) => value.type === blockType))[0],
```
bzabos, line 82: I'm personally not a fan of this type of abstraction, since we're returning a closure, it's going to generate a new function definition every time it's called, it makes it unobvious at the consumer call sites as to what is data and what needs to be invoked, and additionally: this function then needs to scan all block classes in every call on every redraw iteration at the template level where we're calling this from! Depending on how much data we're dealing with, this pattern can get smelly quite quick.

From my perspective this would be happier living as a standalone function at the module level that we can import and use. Otherwise, I'd love to see this rather as an index, like:
```
blockClassesByBlockType: ({ui}, getters, rootState) => keyBy(ui.blockClasses, 'type')

// later ...

const cls = blockClassesByBlockType[type]
```

bzabos (changes requested):
Hi @safydy! It looks like we'll have to re-think how we tackle the problem at hand! Let's book a date to hash that out together.

bzabos (changes requested):
Hi @safydy! It looks like we'll have to re-think how we tackle the problem at hand! Let's book a date to hash that out together.

src/views/InteractionDesigner.vue
```diff
@@ -213,14 +214,12 @@
       async registerBlockTypes() {
         const {blockClasses} = this
 
-        forEach(blockClasses, async ({type}) => {
-          const normalizedType = type.replace('\\', '_')
+        forEach(blockClasses, async ({type}, key) => {
           const typeWithoutSeparators = type.replace(/\\/g, '')
-          const exported = await import(
-            `../components/interaction-designer/block-types/${normalizedType}Block.vue`)
+          const exported = await import(`@/${key}.vue`) // static path prefix such as `@/` is important, as webpack won't accept fully dynamic statement such as import(${key}.vue`)
```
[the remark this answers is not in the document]
safydy-r: Ok, let chat then.

Just a prior thought: 
- the `key` here can contain long path like `components/interaction-designer/block-types/ConsoleIO_PrintBlock` (without `@/`
- but we can force the JSON to have `@/` if we want, then make sure we remove it before making import. eg: `const exported = await import(`@/${key.replace('@/', '')}.vue`)`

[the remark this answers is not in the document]
bzabos: - [ ] @bzabos to spike this out and ensure this is feasible with webpack
- [ ]

src/store/trees/trees.js
bzabos: - [ ] @safydy to remove parameterized getter in favour of either a module-level function or a pre-indexed computed getter over `blockType`.

safydy-r: removed `blockClassFromBlockType` usage

src/store/trees/10-trees-model.js
```diff
@@ -818,7 +818,8 @@ window.app = window.app || {};
 				missingLanguageIdsSocialContent = [];
 				missingLanguageIdsClipboardContent = [];
 
-				if(app.ui.blockClasses[block.type].hasContent == 1){
+				const blockClass = _.pickBy(app.ui.blockClasses, (value, key) => value.type === block.type)
```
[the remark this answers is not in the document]
safydy-r: I think `find()` would work if `app.ui.blockClasses` is collection, but it's an `object`

safydy-r:
We agreed to close this because @bzabos has already included those changes in his PR  (I don't remember the ticket id though :D ).

PR #38, opened 2021-03-11
Vmo 3601 GitHub pages auto rebuild

safydy-r:
To use travis-ci making auto build, I followed this https://cli.vuejs.org/guide/deployment.html#github-pages

I create git patches to deal with the `router` & `vue.config.js` pre-build changes

I have tested the travis-ci config on current branch, see build sample: https://travis-ci.com/github/FLOIP/flow-builder/jobs/489912170
and noticed that it skipped the gh-pages deployment because we're in a PR 
```
Skipping a deployment with the pages provider because the current build is a pull request.
```

I think we need to merge the PR to see it working


.travis.yml
```diff
@@ -4,3 +4,13 @@ node_js:
...
 script:
   - yarn test:unit --no-cache
+  - bash scripts/prepare-gh-pages-deployment.sh
+  - yarn build
+deploy:
+  provider: pages
+  skip_cleanup: true
+  github_token: $GITHUB_TOKEN
+  local_dir: dist
+  target_branch: gh-pages-ci
```
safydy-r, line 14: I want a different branch if it was built from `travis-ci`

.travis.yml
```diff
@@ -1,6 +1,16 @@
...
+  on:
+    branch: VMO-3601-github-pages-auto-rebuild
```
safydy-r, line 16: Will change this to `master` once approved

.travis.yml
```diff
@@ -1,6 +1,16 @@
 language: node_js
 node_js:
   - "stable"
-cache: yarn
+install: yarn install
```
safydy-r, line 4: this fixes an yarn issue https://travis-ci.com/github/FLOIP/flow-builder/builds/219689301#L226
`The command "eval yarn --frozen-lockfile " failed`

scripts/prepare-gh-pages-deployment.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# This apply needed changes before running yarn build,
+# as described in https://github.com/FLOIP/flow-builder/blob/master/gh-pages-deploy.md
+
+echo "Prepare gh-pages deployment"
+git apply patches/gh-pages-prepare-router.patch
```
bzabos, line 7: @safydy-r this is brilliant + elegant to use git patches! 

What happens in the future as we make changes to the affected sources? Will we quite quickly end up with conflicts, or will git handle that intelligently?

bzabos (approved):
This is good! Thanks for the thoughtful solution. It looks like it will be more maintainable than some of the hacks I had in mind😹

scripts/prepare-gh-pages-deployment.sh
safydy-r: Thanks 😊 .
In future, in case on conflict, we will correct the patch file. Let me add some notes in this file

PR #42, opened 2021-03-16
Vmo 3602 add tooltips in canvas for blocks

safydy-r:
[Merge plan]
Waiting for vmo-3369

[UI sample overview]
Screen Shot 2021-03-16 at 2 59 54 PM
Screen Shot 2021-03-18 at 3 27 41 PM


src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -119,6 +119,7 @@
 
           <button v-if="isEditable"
                   type="button"
+                  v-b-tooltip.hover="transIf(activeBlockId,'flow-builder.tooltip-duplicate-block')"
```
bzabos, line 122: Rather than the indirection of `transIf()`, could we use the built-in `disabled` configuration property? Like:
```
:v-b-tooltip.hover={title: trans('flow-builder.tooltip-duplicate-block'), disabled: activeBlockId}"
```

src/components/interaction-designer/Block.vue
```diff
@@ -64,6 +64,7 @@
                              :id="`exit/${exit.uuid}/pseudo-block-handle`"
                              :key="`exit/${exit.uuid}/pseudo-block-handle`"
                              :is-editable="isEditable"
+                             v-b-tooltip.hover.bottom="transIf(isEditable, 'flow-builder.tooltip-new-connection')"
```
bzabos, line 67: Can we bump these exit tooltips to be up on top rather than bottom? 😻 or does it look bad? It feels a bit odd not being able to see the connection coming out the bottom. 

If it came out the top, would we want it to appear above the exit label or the button?

bzabos (approved):
This looks great! 
Approved pending minor suggestions.

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
safydy-r: @bzabos , we're already using the `disabled` prop.
But in this particular scenario, if we delete the block, the delete tooltip remains open.
Screen Shot 2021-03-18 at 2 27 17 PM

Developers have reported this: https://github.com/bootstrap-vue/bootstrap-vue/issues/1892

I will implement the suggested solution then: `Place your tooltip on an outer wrapper element, rather than the disabled button element.` https://github.com/bootstrap-vue/bootstrap-vue/issues/1892#issuecomment-502703464

But this means, the user will still see the tooltip when making a mouse over

safydy-r: But to disable tooltip on block side when we are on **View Mode**, the above solution won't work. Eg: https://github.com/FLOIP/flow-builder/pull/42/files#diff-43f7c5f1a1252fa694baadbe4e2a0ec2f48c3a9041968f25ed1cfab7b4bdff9aR102

I think, I'm keeping the `transIf()` workaround for now, unless you have another solution.

src/components/interaction-designer/Block.vue
safydy-r: Ok, let me just put it on `top` for now

src/components/interaction-designer/Block.vue
```diff
@@ -136,6 +140,9 @@ import { ResourceResolver, SupportedMode } from '@floip/flow-runner'
 import { OperationKind, generateConnectionLayoutKeyFor } from '@/store/builder'
 import Connection from '@/components/interaction-designer/Connection.vue'
 import lang from '@/lib/filters/lang'
+import {VBTooltipPlugin} from 'bootstrap-vue'
+
+Vue.use(VBTooltipPlugin)
```
poojahpatil90, line 145: Minor thing: I would like to understand if there a specific reason why we use a bootstrap-vue plugin inside a specific component? 
Can we do something like this inside main.ts instead?
```
import BootstrapVue from 'bootstrap-vue'
Vue.use(BootstrapVue)
```
This would make all the bootstrap components available across the app.

poojahpatil90 (approved):
This looks great!  I have approve this PR.
I have a comment to understand about the usage of the tooltip component.

src/components/interaction-designer/Block.vue
safydy-r: @poojahpatil90 , for this particular scenario there is no specific reason. I see it could be confusing, so I changed it to
```
import { BTooltip } from 'bootstrap-vue'

Vue.component('b-tooltip', BTooltip)
```
I want to avoid importing all bootstrap-vue components without using them

poojahpatil90: Thanks for clarifying. Just wanted to understand the reasoning.

PR #43, opened 2021-03-17
Vmo 2297 add set group membership block

safydy-r:
[UI Overview]
Screen Shot 2021-03-17 at 12 56 40 PM


src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -0,0 +1,128 @@
...
+  mixins: [lang],
+})
+class Core_SetGroupMembershipBlock extends Vue {
+  @Prop() readonly block!: IBlock
+  @Prop() readonly flow!: IFlow
+
+  actionsList: IGroupOption[] = [
+    {
+      id: ADD_KEY,
+      // @ts-ignore: TODO: fix TS issue `Property 'trans' does not exist on type 'Core_SetGroupMembershipBlock'`
```
safydy-r, line 88: I spent hours to fix this issue but no solution so far. 
I even tried to use class based mixins to import the `lang`, like
```
class Core_SetGroupMembershipBlock extends mixins(Lang) {
...
}
```
in which `Lang` is added & exported from  `@/lib/filters/lang` to match this pattern https://class-component.vuejs.org/guide/extend-and-mixins.html#mixins

If you have an idea to fix them, pls let me know

src/components/interaction-designer/block-editors/BlockId.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="text-right">
+  <div class="block-id text-right">
```
safydy-r, line 2: Note: I took the opportunity to add  missing `class` on existing component to match our convention https://docs.google.com/document/d/1TDnCLixJKGAEIcW2V8prlHDIqrh8UVh-d0bwXOIXX3Q/edit#

poojahpatil90 (approved):
Everything looks good to me, approved!

builder.config.json
```diff
@@ -236,7 +243,16 @@
...
     "groupNames": {},
-    "groups": [],
+    "groups": [
+      {
+        "id": 987,
+        "name": "Group 1"
+      },
+      {
+        "id": 988,
+        "name": "Group 2"
```
bzabos, line 253: Is this sample data? Should we take it out?

src/components/interaction-designer/block-editors/BlockId.vue
bzabos: Thank you! 👌

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    return null
+  }
+
+  set selectedAction(value: IGroupOption) {
+    this.block_updateConfigByPath({
+      blockId: this.block.uuid,
+      path: 'isMember',
+      value: value === null || value === undefined ? null : (value.id === ADD_KEY),
```
bzabos, line 119: Can we push this action into `SetGroupMembershipStore` as `setIsMember(...` so that this domain logic around `value` is encapsulated and reusable?

src/components/interaction-designer/block-editors/GroupSelector.vue
```diff
@@ -0,0 +1,76 @@
...
+      value: value.id
+    })
+    this.block_updateConfigByPath({
+      blockId: this.block.uuid,
+      path: 'groupName',
+      value: value.name
+    })
+  }
+
+  get groupsList(): object[] {
```
bzabos, line 68: Can we rename to `groups` to keep it consistent all the way through?

bzabos (approved):
This looks good, @safydy! I left a couple suggestions, nothing major. 
Thanks!

builder.config.json
safydy-r: hey @bzabos , sometimes I get quite confused when should we add sample data (eg: `languages`, `audio.library`, etc), and when we shouldn't ? Can you give more precision pls ?

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -0,0 +1,128 @@
...
+      name: this.trans('flow-builder.remove'),
+    },
+  ]
+
+  get propertyValue(): string {
+    return get(this.block, 'config.set_contact_property.property_value', '')
+  }
+
+  get selectedAction() {
+    const isMember = get(this.block, 'config.isMember')
```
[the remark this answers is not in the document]
safydy-r: you're right here, thanks for noticing ;)

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -0,0 +1,128 @@
...
+    },
+    {
+      id: REMOVE_KEY,
+      // @ts-ignore: TODO: fix TS issue `Property 'trans' does not exist on type 'Core_SetGroupMembershipBlock'`
+      name: this.trans('flow-builder.remove'),
+    },
+  ]
+
+  get propertyValue(): string {
+    return get(this.block, 'config.set_contact_property.property_value', '')
```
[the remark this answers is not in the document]
safydy-r: Will this one too

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
safydy-r: That's a good idea, moving it into a store `action`

src/components/interaction-designer/block-editors/GroupSelector.vue
safydy-r: sure, thanks

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
safydy-r: I never thought about this solution, thank you very much 💯

builder.config.json
safydy: this thread has been clarified in vmo-2296 PR.
So, I removed this change and created stories instead

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -0,0 +1,125 @@
...
+    },
+    {
+      id: REMOVE_KEY,
+      name: trans('flow-builder.remove'),
+    },
+  ]
+
+  get selectedAction() {
+    const { isMember } = this.block.config as ISetGroupMembershipBlockConfig
+    //TODO: we can remove the safe cast JSON.parse(isMember) once ISetGroupMembershipBlockConfig isMember type is changed to boolean
```
bzabos, line 100: > [...] once ISetGroupMembershipBlockConfig isMember type is changed to boolean

When does this happen?

src/store/flow/block-types/Core_SetGroupMembershipStore.ts
```diff
@@ -0,0 +1,61 @@
...
+      await dispatch('flow/block_createBlockDefaultExitWith', {
+        props: ({
+          uuid: (new IdGeneratorUuidV4()).generate(),
+          tag: 'Default',
+          label: 'Default',
+        }) as IBlockExit,
+      }, { root: true }),
+    ]
+
+    return defaults(props, {
```
bzabos, line 32: It doesn't look obvious here: do you think it matters that `config` property below will always be overwritten in its entirety if provided as part of props argument into this function? 

Like, if defaults has `{config: {groupKey, groupName, isMember}}`, and I provide props as `{config: {isMember: true}}`, then the data returned will have `config` as `{isMember: true}` and without the other properties, because defaults is shallow. 

An option to consider: use `defaultsDeep(…)`

Is it important? 
Does it need to be standardized across `createWith()`s so that behaviour is predictable?

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -0,0 +1,125 @@
...
+
+    if (JSON.parse(isMember) === true) {
+      return find(this.actionsList, { id: ADD_KEY }) || null
+    }
+
+    return null
+  }
+
+  set selectedAction(group: IGroupOption) {
+    this.setIsMemberFromGroup(group)
```
bzabos, line 113: I'm not expecting any changes here, but wanted to take this opportunity to share some food for thought: In the future, can we try to consolidate where "behaviour" is encapsulated? 

In this case we have some special domain constraints that has isMember correlate directly to an action over a groupId, except we have half of the behaviour in vuex and half of it here as the getter. From my perspective, it could be nice to encapsulate both counterparts as part of this computed property so that we don't need to leak UI layer's correlation onto many layers. 

So the vuex action would end up being dirt simple, like: `setIsMember(true | false)` and the translation happens at the same layer for both get and set of action.

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -0,0 +1,125 @@
...
+import SetGroupMembershipStore, { BLOCK_TYPE, ADD_KEY, REMOVE_KEY } from '@/store/flow/block-types/Core_SetGroupMembershipStore'
+import lang, { trans } from '@/lib/filters/lang'
+import { createDefaultBlockTypeInstallerFor } from '@/store/builder'
+import { find } from 'lodash'
+
+const blockVuexNamespace = namespace(`flow/${BLOCK_TYPE}`)
+const flowVuexNamespace = namespace('flow')
+const builderVuexNamespace = namespace('builder')
+
+interface IGroupOption {
```
bzabos, line 64: Can we rename `IGroupOption` to something like `IGroupActionOption` ? 

The difference is then in the shorthand used to reference it which gets misleading. Later we shortened this to `group: IGroupOption `, which then made it into vuex as `group` where the comparison `group.id === ADD_KEY` broke my mind a bit 😹 

If we do the above, can we also rename the variables of that type to be shorthanded as `action` rather than `group`?

Thanks!

bzabos (approved):
Hi @safydy, I've left a couple final thoughts and the only thing I would suggest is a rename:
- [ ] rename IGroupOption to something like IGroupActionOption
- [ ] update to use `defaultsDeep(…)` if it's relevant

I've approved so feel welcome to proceed once you're happy with the PR!

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
safydy: @bzabos , honestly, I don't know who should I contact for flow-runner change request like this. I just notified you with Roan & Bulat on slack https://votomobile.slack.com/archives/CMQDVRDN3/p1616178709010700 
I can create a ticket for it, but I'm not sure I have the full context in case this is already handled in a specific ticket, so I need your opinion here.

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
safydy: Very nice catch. I totally agree here. (I made some copy paste from the GroupSelector.vue and forgot we're having `action` not `group` in this component). Renaming ...

src/store/flow/block-types/Core_SetGroupMembershipStore.ts
safydy: This is a very good point! :100: 
I think we need to standardized across all `createWith()`, as both of us understand the impact and are in the same page, I will make changes right in this PR to avoid making new ticket.

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
safydy-r: Thanks for explaining, noting this ;)

PR #55, opened 2021-03-25
Vmo 2623 tidy new code to remove ts errors

safydy-r:
[Remaining // ts-ignore in `src/` folder]
- in TreeBuilderToolbar.vue & flow.ts: related with IBlock `platform_metadata`, I think it's ok for now

For dev info:
I found a solution to use mixins with vue-class based components:
- it's important to turn mixins files into `.ts` first (I don't know why js files won't work)
- then follow this pattern https://class-component.vuejs.org/guide/extend-and-mixins.html


src/components/interaction-designer/resource-editors/ResourceEditor.vue
```diff
@@ -113,40 +115,27 @@ const builderVuexNamespace = namespace('builder')
     created_at: string;
   }
 
-  @Component({
-    props: {
```
safydy-r, line 117: turned to vue-class based

src/components/interaction-designer/resource-editors/ResourceEditor.vue
```diff
@@ -113,40 +115,27 @@ const builderVuexNamespace = namespace('builder')
...
+  @Component({
     components: {
       AudioLibrarySelector,
       ResourceVariantTextEditor,
       UploadMonitor,
       PhoneRecorder,
     },
   })
-export class ResourceEditor extends Vue {
+export class ResourceEditor extends mixins(FlowUploader, Permissions, Routes, Lang) {
```
safydy-r, line 130: Note this guys, you may need it for future usage ;)

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -172,234 +169,266 @@ import pickBy from 'lodash/fp/pickBy'
...
+const flowVuexNamespace = namespace('flow')
+const builderVuexNamespace = namespace('builder')
+
+@Component({
   components: {
     // Affix,
     // TreeUpdateConflictModal,
     // InteractionTotalsDateRangeConfiguration
   },
-  mixins: [
```
safydy-r, line 184: turned into vue-class based

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
```diff
@@ -51,31 +51,37 @@ export const getters: GetterTree<IFlowsState, IRootState> = {
 }
 
 export const mutations: MutationTree<IFlowsState> = {
-  deleteChoiceByKey(state, { choiceKeyToRemove, blockId }) {
```
safydy-r, line 54: As we don't have access to rootState  from Mutation & this.state.flow throws ts exception, I turned this into Actions

src/components/interaction-designer/block-types/ConsoleIO_PrintBlock.vue
```diff
@@ -33,30 +33,30 @@ import { IPrintBlock } from '@floip/flow-runner/src/model/block/IPrintBlock'
...
+  components: {
+    ResourceEditor,
+    BlockNameEditor,
+    BlockLabelEditor,
+    BlockSemanticLabelEditor,
+    FirstBlockEditorButton,
+    BlockId,
+  },
+})
+class ConsoleIO_PrintBlock extends mixins(Lang) {
```
jcbashdown, line 59: Nice. I almost did this in my branch but decided not to mess with `@/lib/filters/lang`. Good to finally get it done. I notice a few places we will use the old `mixins: [lang]`. Should we update those?

src/store/flow/block.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '@floip/flow-runner'
 import { ActionTree, GetterTree, MutationTree } from 'vuex'
 import { IRootState } from '@/store'
-import { defaults, without } from 'lodash'
+import { defaults, set } from 'lodash'
```
jcbashdown, line 12: Noticed you also do this in 2296. I suppose this will shake out in the conflict resolution...

src/store/flow/flow.ts
```diff
@@ -77,9 +77,6 @@ export const mutations: MutationTree<IFlowsState> = {
       const exitsTowardUs = exits.filter((e) => e.destinationBlock === blockId)
       forEach(exitsTowardUs, (e) => e.destinationBlock = undefined)
     })
-
-    // @ts-ignore
-    this.state.builder.activeBlockId = null
```
jcbashdown, line 82: How come this is no longer necessary?

tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "allowSyntheticDefaultImports": true,
     "sourceMap": true,
     "baseUrl": ".",
+    "strictPropertyInitialization": false,
```
jcbashdown, line 14: What effect does this have exactly?

jcbashdown (changes requested):
Looks good but I have some concerns about merge order and I've made a couple of comments about small possible issues/changes. Let me know what you think.

src/components/interaction-designer/block-types/ConsoleIO_PrintBlock.vue
safydy: Changing the old approach `mixins: [lang]` requires we turn the component into full vue class, which could be time consuming and is not very important for now. I'd propose we create a ticket for Beta version to turn all component into class based.

src/store/flow/block.ts
safydy: Thanks for noticing. I think it will be a conflict. Anyway, any change in this line would create a conflict, so I will leave this here.

src/store/flow/flow.ts
safydy: Um, IFlowState doesn't have `builder` prop, therefore this is always a TS error. So I moved this out of this mutation https://github.com/FLOIP/flow-builder/pull/55/files#diff-3db5c56652da6d4f7b33c23e432a38b9b73d0b30d47df648929798e4c82d498bR334.

But when thinking longer, I found a solution:
- instead of having a mutation, which doesn't give us access to `commit` or `dispatch`
- I changed it into action. Then I changed this line into `commit('builder/activateBlock', { blockId: null }, { root: true })`

Thanks for noticing

tsconfig.json
safydy: Well, this is more like a proposal than a must have change.

here is the definition from the official doc https://www.typescriptlang.org/tsconfig#strictPropertyInitialization
> TypeScript will raise an error when a class property was declared but not set in the constructor.

Why do we need it?
Our current code has TS error like this
```
ERROR in /home/safydy/Documents/flow-builder/src/components/interaction-designer/resource-editors/ResourceEditor.vue(131,29):
131:29 Property 'block' has no initializer and is not definitely assigned in the constructor.
    129 |   })
    130 | export class ResourceEditor extends mixins(FlowUploader, Permissions, Routes, Lang) {
  > 131 |   @Prop({ required: true }) block: IBlock
           |                                            ^
    132 | 
    133 |   @Prop({ required: true }) flow: IFlow
    134 | 
ERROR in /home/safydy/Documents/flow-builder/src/components/interaction-designer/resource-editors/ResourceEditor.vue(133,29):
133:29 Property 'flow' has no initializer and is not definitely assigned in the constructor.
    131 |   @Prop({ required: true }) block: IBlock
    132 | 
  > 133 |   @Prop({ required: true }) flow: IFlow
           |                                            ^
    134 | 
    135 |   @Prop() resource: IResourceDefinition
```

The solution is to provide initial value `@Prop({ required: true, default: {} }) block: IBlock = {} as IBlock`, but it's too repetitive to my POV. 
As I never had an issue without having initialized a prop in the past, I thought we could ignore this kind of error from the tsconfig. What do you think?

src/store/flow/flow.ts
safydy: put it back https://github.com/FLOIP/flow-builder/pull/55/files#diff-eea673c657b7edcdbf00ab176fcc8ec0a84d21e8a8dfa9ff8a014c7510a144a0R134

src/components/interaction-designer/block-types/ConsoleIO_PrintBlock.vue
jcbashdown: Sounds good. Let's leave it

src/store/flow/flow.ts
jcbashdown: Okay, cool. I think that should work

tsconfig.json
jcbashdown: How about this instead? 

https://github.com/kaorun343/vue-property-decorator/issues/81#issuecomment-365941684

The issue with this approach is that I think it is hiding that resource and label are not required/don't have a default. We should make sure we set these in all cases.

jcbashdown: BTW. I suggest we don't prioritise this ticket for the next sprint while we still have functional changes to do. We won't be able to merge until flow lifecycle is merged anyway.

safydy: @jcbashdown , thanks for the suggestion. Let's adopt the `definite assignment assertion` then. I have remove the tsconfig `"strictPropertyInitialization": false`.
About priority, yes we should prioritize flow lifecycle as discussed, but upcoming functionalities (eg: Validation API tickets: vmo-3710, vmo-3704, etc) were branched out of this to avoid headache on conflict resolution. Our plan is to merge this asap once flow lifecycle is merged.

jcbashdown (approved):
Looks good! Hoping to merge lifecycle this morning and unblock finishing this

src/store/flow/block-types/Core_RunFlowBlockStore.ts
```diff
@@ -11,14 +12,11 @@ import { IFlowsState } from '../index'
...
-    // TODO - this should actually be container.flows?
-    // TODO - why does this error in typescript? - flow does exist on IRootState etc...
-    // @ts-ignore - TS2339: Property 'flow' does not exist on type
-    rootState.flow.flows.filter((flow: IFlowsState) =>
-      // @ts-ignore - TS2339: Property 'flow' does not exist on type
-      flow.uuid !== rootGetters['flow/activeFlow'].uuid)
-  ,
+  otherFlows: (state, getters, rootState, rootGetters): IFlow[] => {
+    return rootState.flow.flows.filter((flow: IFlow) => {
+      return flow.uuid !== rootGetters['flow/activeFlow'].uuid
```
bzabos, line 17: Interesting! How come this syntax worked and the rocket (`=>`) shorthand wouldn't? 🙀

bzabos (approved):
Wow, thanks! Looks good to me.

src/store/flow/block-types/Core_RunFlowBlockStore.ts
safydy-r: No idea, sorry :D

PR #56, opened 2021-03-25
Vmo 2623 tidy new code to remove ts errors  storybook part

safydy-r:
to avoid a giant PR, I separated here the storybook changes

[Remaining // ts-ignore type in `stories/` folder]
1 : in all block-type stories we have something like
```
// @ts-ignore
await baseMounted.bind(this)(BLOCK_TYPE, caseBlockStore)
```

eg2:

### DIFF
--- a/stories/CaseBlockStyled.stories.ts
+++ b/stories/CaseBlockStyled.stories.ts
@@ -10,7 +10,6 @@ import caseBlockStore, {BLOCK_TYPE} from '@/store/flow/block-types/Core_CaseBloc
 
 import { baseMounted, BaseMountedVueClass } from './story-utils/storeSetup'
 import {namespace} from "vuex-class";
-const flowVuexNamespace = namespace('flow')
 const blockVuexNamespace = namespace(`flow/${BLOCK_TYPE}`)
 
 Vue.use(Vuex)
@@ -55,9 +54,7 @@ export const Default = () => (DefaultClass)
     const { block } = await baseMounted.bind(this)(BLOCK_TYPE, caseBlockStore)
     const blockId = block.uuid
 
-    this.block_setName({blockId: blockId, value: "A Name"})
-    this.block_setLabel({blockId: blockId, value: "A Label"})
-    this.block_setSemanticLabel({blockId: blockId, value: "A Semantic Label"})
+    this.setDescription(blockId)
     this.editCaseBlockExit({identifier: block.exits[0].uuid, value: "A expression"})
   }
 })
--- a/stories/CaseBlockStyled.stories.ts
+++ b/stories/CaseBlockStyled.stories.ts
@@ -34,7 +34,7 @@ const BaseOptions = {
 }
 
 // default case block state
-@Component<any>({
+@Component({
   ...BaseOptions,
   store: new Vuex.Store<IRootState>(store),
   async mounted() {
@@ -46,7 +46,7 @@ class DefaultClass extends BaseMountedVueClass {}
 export const Default = () => (DefaultClass)
 
 //ExistingDataBlock
-@Component<any>({
+@Component({
   ...BaseOptions,
   store: new Vuex.Store<IRootState>(store),
   async mounted() {
--- a/stories/CaseBlockStyled.stories.ts
+++ b/stories/CaseBlockStyled.stories.ts
@@ -55,21 +55,13 @@ export const Default = () => (DefaultClass)
     const { block } = await baseMounted.bind(this)(BLOCK_TYPE, caseBlockStore)
     const blockId = block.uuid
 
-    // @ts-ignore - TS2339: Property 'block_setName' does not exist on type
     this.block_setName({blockId: blockId, value: "A Name"})
-    // @ts-ignore - TS2339: Property 'block_setLabel' does not exist on type
     this.block_setLabel({blockId: blockId, value: "A Label"})
-    // @ts-ignore - TS2339: Property 'block_setSemanticLabel' does not exist on type
     this.block_setSemanticLabel({blockId: blockId, value: "A Semantic Label"})
-    // @ts-ignore - TS2339: Property 'editCaseBlockExit' does not exist on type
     this.editCaseBlockExit({identifier: block.exits[0].uuid, value: "A expression"})
   }
 })
 class CurrentClass2 extends BaseMountedVueClass {
   @blockVuexNamespace.Action editCaseBlockExit:any
-
-  @flowVuexNamespace.Mutation block_setName:any
-  @flowVuexNamespace.Mutation block_setLabel:any
-  @flowVuexNamespace.Mutation block_setSemanticLabel:any
 }
 export const ExistingDataBlock = () => (CurrentClass2)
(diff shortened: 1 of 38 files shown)


stories/CaseBlockStyled.stories.ts
```diff
@@ -32,44 +32,32 @@ const CaseBlockTemplate = `
...
     const blockId = block.uuid
 
-    // @ts-ignore - TS2339: Property 'block_setName' does not exist on type
-    this.block_setName({blockId: blockId, value: "A Name"})
-    // @ts-ignore - TS2339: Property 'block_setLabel' does not exist on type
-    this.block_setLabel({blockId: blockId, value: "A Label"})
-    // @ts-ignore - TS2339: Property 'block_setSemanticLabel' does not exist on type
-    this.block_setSemanticLabel({blockId: blockId, value: "A Semantic Label"})
-    // @ts-ignore - TS2339: Property 'editCaseBlockExit' does not exist on type
+    this.setDescription(blockId)
```
safydy-r, line 56: As the vue-class extends is working now, I moved repeated code into `story-utils/storeSetup.ts`'s classes

stories/story-utils/storeSetup.ts
```diff
@@ -1,61 +1,133 @@
...
- * Safe register block module
- * Because some weird race condition is leading to modules not getting unregistered when clicking between stories before the next story re-registers
+ * Vue class used to gather required Getter, Mutation, Action for the BaseMounted binding
  */
-export const safeRegisterBlockModule = async function(BLOCK_TYPE: string, blockTypeStore: IRootState) {
-  // @ts-ignore - TS2551: Property '$store' does not exist on type
-  if (this.$store.hasModule(['flow', BLOCK_TYPE])) {
-    // @ts-ignore - TS2551: Property '$store' does not exist on type
-    this.$store.unregisterModule(['flow', BLOCK_TYPE])
+@Component<any>({})
```
safydy-r, line 22: @jcbashdown , If you remember we couldn't extends class correctly, this was the solution. We have to add `@Component` decorator for Parent & Child classes

tests/unit/__snapshots__/storybook.spec.ts.snap
```diff
@@ -14,7 +14,156 @@ exports[`Storyshots ConsoleIo/Print Block Existing Data Block 1`] = `
...
+                         
+                        <div
+                          class="text-right"
+                        >
+                          <small
+                            class="text-muted"
+                          >
+                            Block Id: 
+                            <em>
+                              6f67fcec-c5ba-4451-ae76-51a29b6b30fd
```
safydy-r, line 152: @jcbashdown & @bzabos , it seems like the storyshots is back (before we had empty `<div>`). I don't know what resolved it. But we're having a pbm, for instance, in this particular line, the Uuid in snapshot doesn't match with Uuid during `yarn test:unit`, which will lead to snapshot error.
See current build example: https://travis-ci.com/github/FLOIP/flow-builder/builds/221253135#L352

Any idea how should we handle this ?

stories/CaseBlockStyled.stories.ts
```diff
@@ -1,16 +1,16 @@
 import Vue from 'vue'
 import Vuex from 'vuex'
-import {Component} from 'vue-property-decorator'
+// import {Component, Vue} from 'vue-property-decorator'
```
safydy-r, line 3: will clean this with PR feedback

stories/story-utils/storeSetup.ts
jcbashdown: Do you remember exactly what the issue was? I have extended a flow-builder class with a custom template in voto5 but had some weird behaviour. I had to define a proxy between activeFlow on the child and the parent as even though I could see the computed prop on the child in vue devtools I couldn't actually access the value correctly:

https://bitbucket.org/voto/voto5/pull-requests/1787/vmo-3120-flow-lifecycle#Lresources/assets/js/floip/builder/views/CustomNewFlow.vueT47

tests/unit/__snapshots__/storybook.spec.ts.snap
jcbashdown: The uuid generator gives us something new every run. Check this out:

https://github.com/FLOIP/flow-builder/pull/35/files#diff-48aa2e4c8dec9b1a411f6d75bd7208f6e7ff39633b413f167b397aac9319b477R52

But that approach won't work here or will be time consuming to do. We'll need to fake the uuid generator! We'll have to refactor so this is injected or overidden. Should be doable, could you look into it?

I notice there are also some other errors in the build - both in travis and locally for me I see things like:

`"TypeError: Cannot read property 'reduce' of undefined"`

I suspect this is something I saw in my branch which I resolved in the cases I found it by having lots of v-if conditions in my stories:

https://github.com/FLOIP/flow-builder/pull/35/files#diff-e3f6b63837f0b7bef6c0bfd5de701761ba00d57af7911ed3e58b87735e0772bcR27

jcbashdown:
Going to hold off on a full review of this for now until the build is working again.

tests/unit/__snapshots__/storybook.spec.ts.snap
jcbashdown: p.s. Really good news the sidebar storyshots are back though!

stories/story-utils/storeSetup.ts
safydy: @jcbashdown : the pbm was we couldn't access correclty the parent's props & methods from its children.

Thanks to StackOverFlow, someone has replied to my question and it seems working :+1:
https://stackoverflow.com/questions/63004313/vue-class-component-inheritance-method-from-base-class-is-not-a-function/66194360#66194360

tests/unit/__snapshots__/storybook.spec.ts.snap
safydy: thank for the idea, I will check how we can mock the uuid during tests

src/store/builder/index.ts
```diff
@@ -63,7 +63,7 @@ export interface IBuilderState {
 
 export const stateFactory = (): IBuilderState => ({
   activeBlockId: null,
-  isEditable: false,
+  isEditable: true,
```
safydy, line 66: made `builder` editable as discussed on slack

src/components/common/PlainDraggable.vue
```diff
@@ -69,7 +69,7 @@ export default {
       // prevent css translate() animations for move
       // they don't seem to be throttled enough for leaderline to follow tightly
       leftTop: false,
-      disabled: true,
+      disabled: false, // synced with src/store/builder/index.ts:isEditable=true
```
safydy, line 72: as we made `builder` editable according to what we discussed on slack

stories/story-utils/storeSetup.ts
jcbashdown: Only just realised you asked the question. Nice!

jcbashdown (approved):

bzabos (approved):
Looks good to me! 

TBH, I haven't gotten my hands too deep into storybook lately, hoping I can lean on you two for this one!

PR #59, opened 2021-03-30
VMO-2622 - probably not smart idea on how to clean up resources.

jcbashdown:


src/store/flow/flow.ts
```diff
@@ -171,7 +171,15 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
   async flow_persist({ state, getters, commit }, { persistRoute, flowContainer }): Promise<IContext> {
     const restVerb = flowContainer.created ? 'put' : 'post'
     const oldCreatedState = flowContainer.created
-    return axios[restVerb](persistRoute, omit(flowContainer, ['created']))
+
+    const flowContainerCleanedResources = cloneDeep(flowContainer)
+    flowContainerCleanedResources.resources = []
+    flowContainer.resources.forEach((resource, i) => {
+      if(JSON.stringify(flowContainer.flows).includes(resource.uuid)) {
```
bzabos, line 178: Can we pull the JSONification out of the loop to mitigate slow-by-slow looping? 😛

src/store/flow/flow.ts
```diff
@@ -171,7 +171,15 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
   async flow_persist({ state, getters, commit }, { persistRoute, flowContainer }): Promise<IContext> {
     const restVerb = flowContainer.created ? 'put' : 'post'
     const oldCreatedState = flowContainer.created
-    return axios[restVerb](persistRoute, omit(flowContainer, ['created']))
+
+    const flowContainerCleanedResources = cloneDeep(flowContainer)
```
bzabos, line 175: To separate concerns of interrogation and mutation, we could potentially do without the heavier `cloneDeep()`, favouring collecting a list of resources in this first pass, then afterward rewrite the same `resources` property, maybe something like:
```
flowContainer.resources = without(flowContainer.resources, orphanedResources)
```

src/store/flow/flow.ts
```diff
@@ -181,7 +189,7 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
         commit('flow_updateCreatedState', oldCreatedState)
         if(!persistRoute) {
           console.info("Flow persistence route not configured correctly in builder.config.json. Falling back to vuex store")
-          commit('flow_updateFlowContainer', flowContainer)
+          commit('flow_updateFlowContainer', flowContainerCleanedResources)
           return getters.activeFlowContainer 
         }
         return null
```
bzabos, line 187: Can we always log something in our catch blocks to avoid silently swallowing unexpected errors?

src/store/flow/flow.ts
```diff
@@ -171,7 +171,15 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
-    return axios[restVerb](persistRoute, omit(flowContainer, ['created']))
+
+    const flowContainerCleanedResources = cloneDeep(flowContainer)
+    flowContainerCleanedResources.resources = []
+    flowContainer.resources.forEach((resource, i) => {
+      if(JSON.stringify(flowContainer.flows).includes(resource.uuid)) {
+        flowContainerCleanedResources.resources.push(resource)
+      }
+    })
+    return axios[restVerb](persistRoute, omit(flowContainerCleanedResources, ['created']))
```
bzabos, line 182: From another changeset: I'd love to see this use asynchronous/await! 🤩

bzabos (approved):
Hi @jcbashdown, I think this is actually quite elegant in its simplicity! Thanks for the creative thinking. 

Are you happy with it? Should we create a task to dive into a more intricate solution in the future?

safydy-r:
closing it as found a way in https://viamoinc.atlassian.net/browse/VMO-6643

PR #61, opened 2021-04-07
Vmo 3710 validation api  add vuex store for messaging

safydy:
For now, I put generated jsonSchema from flow-runner into a `_tmp` folder. I will remove them once we consume the right flow-runner version.

And the way we can check validation is on console for now, eg:
```
flow validation status  | isValid: false  | errors: should NOT have additional properties  | details: 
>Object { isValid: Getter & Setter, errors: Getter & Setter, … }

validation status for block 6c7679e5-3bb2-4447-8373-2d8be3657375  | isValid: false  | errors: should NOT have additional properties  | details: 
>Object { isValid: false, errors: (1) […] }
```
TODO
- [x] point to the right JSON once we consume the right flow-runner version, then delete tmp file
- [x] https://github.com/FLOIP/flow-builder/pull/61/files#diff-43838ea39d354dcd7d8e1245d202064a03d4bf33f2bb933b0374d57fef9a0299R14


src/store/validation/index.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export const actions: ActionTree<IValidationState, IRootState> = {
+  async validate_block({ state }, { block } : { block: IBlock }): Promise<IValidationStatus> {
+    const { uuid: blockId, type: blockType } = block
+    const blockTypeWithoutNameSpace = blockType.split('.')[blockType.split('.').length - 1]
+    const blockJsonSchemaFile = `I${blockTypeWithoutNameSpace}Block.json`
+
+    if (isEmpty(state.blockTypeValidators) || !state.blockTypeValidators.hasOwnProperty(blockTypeWithoutNameSpace)) {
+      // TODO: point to the right JSON once we consume the right flow-runner version, then delete tmp file
```
safydy, line 46: Added this TODO as reminder

src/store/validation/index.ts
```diff
@@ -0,0 +1,105 @@
...
+ * const isValid = validate(myData)
+ * const error = validate.errors
+ *
+ * @param jsonSchema
+ */
+export function createDefaultJsonSchemaValidatorFactoryFor(jsonSchema: JSONSchema7): ValidateFunction {
+  return ajv.compile(jsonSchema)
+}
+
+export function debutValidationStatus(status: IValidationStatus, customMessage: string) {
```
safydy, line 98: Will fix typo `debug...` not `debut...`

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -12,24 +12,63 @@
...
+        }
+
+        await this.validate_block({ block: currentNewBlock });
+        debutValidationStatus(this.blockValidationStatuses[currentNewBlock.uuid], `validation status for block ${currentNewBlock.uuid}`)
+      }
+    }
+  }
+
+  get blocksOnActiveFlowForWatcher() {
+    return cloneDeep(this.activeFlow.blocks) // needed to make comparison between new & old values on watcher
```
safydy, line 61: We can improve performance here to not watch `vendor_metadata` changes (eg: delete it before cloning). I tried but it didn't work yet, it may require extra debug & tests. I'd suggest we move the all watching improvements in Beta version.

@bzabos , any thoughts ?

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -12,24 +12,63 @@
...
+
+  @Watch('blocksOnActiveFlowForWatcher', { deep: true, immediate: true })
+  async onBlocksInActiveFlowChanged(newBlocks: IBlock[], oldBlocks: IBlock[]) {
+    if (!!newBlocks && newBlocks.length > 0) {
+      console.debug('blocks inside active flow have changed, validating ...');
+      for (let i = 0;  i < newBlocks.length; i++) {
+        const currentNewBlock = newBlocks[i]
+        const currentOldBlock = find(oldBlocks, { 'uuid': currentNewBlock.uuid })
+
+        if (isEqual(currentNewBlock, currentOldBlock)) {
```
safydy, line 50: Added this to not validate the whole elements inside `flow.blocks[]`

src/store/validation/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  console.debug(
+    customMessage,
+    ' | isValid:', status.isValid,
+    ' | errors:', `${status.hasOwnProperty('errors') && !!status.errors! ? (status.errors!).map(item => get(item, 'message', 'undefined')).join(';') : 'undefined'}`,
```
safydy, line 102: Another proposal, I'd suggest we implement i18n in Beta version https://ajv.js.org/api.html#errors-i18n

If you're ok @bzabos , I can create a ticket for that

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -12,21 +12,54 @@
...
+  },
+})
 export default class BuilderCanvas extends Vue {
+  @Watch('activeFlow', { deep: true, immediate: true })
+  async onActiveFlowChanged(newFlow: IFlow) {
+    console.debug('active flow has changed, validating ...', newFlow)
+    await this.validate_flow({ flow: newFlow })
+  }
+
+  @Watch('blocksOnActiveFlowForWatcher', { deep: true, immediate: true })
```
bzabos, line 37: Would we be able to avoid the indirection of the computed getter (`blocksOnActiveFlowForWatcher`) by using  `activeFlow.blocks` as the string here? It looks like vue supports nesting like that via: https://vuejs.org/v2/api/#vm-watch

I think I'm misunderstanding how come we need the computed getter and cloning :) Would you be willing to help me understand?

bzabos (changes requested):
Hi @safydy, this looks really good! I've added a few comments, the I'm hoping to see is "moving validators out of vuex". 

I'd also love to connect to understand how come we need to clone the blocks in the computed getter for watching. Let me know if you can meet about it.

src/store/validation/index.ts
bzabos: I think this is a great idea! Thanks @safydy !

src/components/interaction-designer/BuilderCanvas.vue
bzabos: I think this would speed up cloning, but I'm curious whether or not it would speed up validation if there are not any rules to validate against that data? I'd be slightly worried that a future/vendor block type could attempt to validate a property within `vendor_metadata` and be prevented from doing so! Do you think that's a fair concern?

src/components/interaction-designer/BuilderCanvas.vue
bzabos: I'm thinking about performance— We're adding overhead with validation, so we'll need to make sure we do some profiling and testing with larger trees after this feature is done to test things out :)

src/store/validation/index.ts
safydy: Great ! created the ticket https://viamoinc.atlassian.net/browse/VMO-3957

src/store/validation/index.ts
```diff
@@ -0,0 +1,157 @@
...
+interface IIndexedString { [key: string]: string }
+
+export interface IValidationStatus {
+  isValid: boolean | PromiseLike<any>;
+  ajvErrors?: null | Array<ErrorObject>;
+}
+
+export interface IValidationState {
+  validationStatuses: { [key: string]: IValidationStatus }; //important context for future debug or testing, keys are index like `flow/flowId`
+  validators: { [key: string]: ValidateFunction }; //AJV validators, keys are types
```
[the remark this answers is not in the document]
safydy: Ok, thanks for letting me know.
I moved `validators` outside the vuex state

I liked the `Map()` object, I didn't use to it. Very handy ❤️  💯

src/store/validation/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  async validate_block({ state, commit }, { block } : { block: IBlock }): Promise<IValidationStatus> {
+    const { uuid: blockId, type: blockType } = block
+    const blockTypeWithoutNameSpace = blockType.split('.')[blockType.split('.').length - 1]
+    const blockJsonSchemaFile = `I${blockTypeWithoutNameSpace}Block.json`
+
+    if (isEmpty(state.validators) || !state.validators.hasOwnProperty(blockTypeWithoutNameSpace)) {
+      // TODO: point to the right JSON once we consume the right flow-runner version, then delete tmp file
+      state.validators[blockTypeWithoutNameSpace] = createDefaultJsonSchemaValidatorFactoryFor(require(`../../../_tmp/${blockJsonSchemaFile}`))
+    }
+    const validate = state.validators[blockTypeWithoutNameSpace]
```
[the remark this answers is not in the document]
safydy-r: Good idea, done !

src/store/validation/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    return state.validationStatuses[index]
+  },
+
+  async validate_flow({ state, commit }, { flow } : { flow: IFlow }): Promise<IValidationStatus> {
+    const validationType = 'flow'
+    if (isEmpty(state.validators) || !state.validators.hasOwnProperty(validationType)) {
+      // TODO: point to the right JSON once we consume the right flow-runner version, then delete tmp file
+      state.validators[validationType] = createDefaultJsonSchemaValidatorFactoryFor(require('../../../_tmp/flowSpecJsonSchema.json'), '#/definitions/IFlow')
+    }
+    const validate = state.validators[validationType]
```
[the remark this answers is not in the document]
safydy-r: Done too

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: @bzabos : 
We want to watch a state here (activeFlow.blocks), but in addition we want to get correct `old` & `new` new values for comparison.
If I watch only `activeFlow.blocks`, I couldn't have the old & new values correctly. They always have the same value 😞  . The only approach I found so far is to put the state under a getter & make cloneDeep 💡 . (check this: https://stackoverflow.com/questions/50523609/how-to-retrieve-the-old-and-new-values-of-a-store-state)
I would be happy if you have another idea to avoid this approach 😄 

For now, there are 02 solutions:
- solution1: comparing new & old values, we watch the getter `blocksOnActiveFlowForWatcher` which makes `cloneDeep` of `activeFlow.blocks`
- solution2: Without comparing new & old, we just need to watch over `activeFlow.blocks`

I tried to make simple profiling in the watcher
```
  @Watch('...', { deep: true, immediate: true })
  async onBlocksInActiveFlowChanged(newBlocks: IBlock[], oldBlocks: IBlock[]) {
    const t0 = performance.now()
      ...
    const t1 = performance.now()
    console.log("Profiling ", t1 - t0, 'ms')
  }
```

The result show us the solution1 has a significant advantage, check profiling result here https://docs.google.com/document/d/1xwMCxm0ilFExehaNMq8SmvrTuVFZvaHJVwrttozVHIE/edit?usp=sharing

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: Hum, you make a point here. Let's not delete the  `vendor_metadata` (if we decide to keep this getters, see later below my explanation of why I put `cloneDeep`)

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -12,21 +12,54 @@
...
+  async onBlocksInActiveFlowChanged(newBlocks: IBlock[], oldBlocks: IBlock[]) {
+    if (!!newBlocks && newBlocks.length > 0) {
```
bzabos, line 39: Due to typing, we shouldn't be able to get into the state where we need this first presence guard against a null `newBlocks`, should we? (eg. `!!newBlocks`)

src/store/validation/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * Human readable errors, keys are index like `flow/flowId/.path/.to/.prop`.
+   * Note that indexedErrors has more elements than validationStatuses.
+   */
+  flattenErrorMessages(state): IIndexedString {
+    let accumulator: IIndexedString = {}
+    forIn(state.validationStatuses, function (validationStatus: IValidationStatus, index: string) {
+      flatValidationStatuses({
+        keyPrefix: index,
+        errors: validationStatus.ajvErrors,
```
bzabos, line 44: Can you drop a sample ajvError in the docblock above?  Eg. `* @example - {...}`

src/store/validation/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  validationStatuses: { [key: string]: IValidationStatus }; //important context for future debug or testing, keys are index like `flow/flowId
+}
+
+export const stateFactory = (): IValidationState => ({
+  validationStatuses: {} as { [key:string]: IValidationStatus },
+})
+
+export const getters: GetterTree<IValidationState, IRootState> = {
+  /**
+   * Human readable errors, keys are index like `flow/flowId/.path/.to/.prop`.
```
bzabos, line 36: Hi hi, are the slashes truly present? I am under the impression that it should be more like: 
```
flow/{flowId}/.path.to.prop
```

The above should then match our deep-linking tasks. Can you confirm?

src/components/interaction-designer/BuilderCanvas.vue
bzabos: Low priority: Also, can we flip this to be a bail-early situation to lay patterns to avoid a Flock of Geese situation? 
```
if (newblocks.length === 0) {
  return
}
```


Flock of Geese situation:
```
>
    >
        >
            >
                >
                    >
                >
            >
        >
    >
>
```

image


😹

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -12,21 +12,54 @@
...
+      console.debug('blocks inside active flow have changed, validating ...');
+      for (let i = 0;  i < newBlocks.length; i++) {
+        const currentNewBlock = newBlocks[i]
+        const currentOldBlock = find(oldBlocks, { 'uuid': currentNewBlock.uuid })
+
+        if (isEqual(currentNewBlock, currentOldBlock)) {
+          continue; // no changes found
+        }
+
+        await this.validate_block({ block: currentNewBlock });
```
bzabos, line 49: I'm seeing a lot of semicolons! Is our standard to be semicolon-less?

src/store/validation/index.ts
```diff
@@ -0,0 +1,165 @@
...
+      throw new Error(`Cannot find definition ${subSchema} in schema ${jsonSchema}`)
+    }
+    return validate as ValidateFunction
+  }
+}
+
+function debugValidationStatus(status: IValidationStatus, customMessage: string) {
+  if (status) {
+    console.debug(
+      customMessage,
```
bzabos, line 139: Can we try to create a pattern of using the first param to `console.*()` methods as the module path? This will help keep our log sane. Eg:
```
console.debug('store/validation', 'Uh oh! The thing happened again…')
```

src/store/validation/index.ts
```diff
@@ -0,0 +1,165 @@
...
+    validators.set(blockType, createDefaultJsonSchemaValidatorFactoryFor(require(`../../../_tmp/${blockJsonSchemaFile}`)))
+  }
+  return validators.get(blockType)!
+}
+
+function getOrCreateFlowValidator(): ValidateFunction {
+  const validationType = 'flow'
+  if (isEmpty(validators) || !validators.has(validationType)) {
+    // TODO: point to the right JSON once we consume the right flow-runner version, then delete tmp file
+    validators.set(validationType, createDefaultJsonSchemaValidatorFactoryFor(require('../../../_tmp/flowSpecJsonSchema.json'), '#/definitions/IFlow'))
```
bzabos, line 108: @safydy Do we have a JIRA for the phase that will account for pulling these in from @floip/flow-runner repo? Did I already ask this? 😹

bzabos (approved):
Hi @safydy! This is great! I added some maintenance notes, but I'm happy with this if you are😸

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: Nice one, thanks for the image 😻 , doing some changes here

src/store/validation/index.ts
safydy-r: Ok , added it

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: Thanks, had new IDE settings on Ubuntu (spare machine) 😆 , removing all

src/store/validation/index.ts
safydy-r: My plan is to merge vmo-3702 (Flow-runner side), then edit this part before sending to QA

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -12,21 +12,56 @@
...
 
 <script lang="ts">
-import { mapGetters } from 'vuex'
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Vue, Watch } from 'vue-property-decorator'
 import Block from '@/components/interaction-designer/Block.vue'
+import { namespace } from 'vuex-class'
+import {IBlock, IFlow} from '@floip/flow-runner'
+import { find, isEqual, cloneDeep } from 'lodash'
+import { IValidationStatus } from '@/store/validation'
```
poojahpatil90, line 20: @safydy-r  since we changed ValidationStatus to ValidStatus, can we change the interface name as well here?

yeboaasare (approved):

PR #63, opened 2021-04-13
Vmo 3704 add validation message component

safydy-r:
I added a test for `flow.interaction-timeout`
image


src/components/interaction-designer/flow-editors/FlowEditor.vue
```diff
@@ -8,7 +8,11 @@
       <div class="row">
           <div :class="{'col-12': sidebar, 'col-6': !sidebar}">
             <flow-label-editor :flow="flow" />
-            <flow-interaction-timeout-editor :flow="flow" />
+            <validation-message
+              :message-key="`flow/${flow.uuid}/.interaction_timeout`"
+              #input-control="{ isValid, test }">
```
safydy-r, line 13: will remove this `test` variable

src/components/common/NumericEditor.vue
```diff
@@ -6,6 +6,7 @@
           type="number"
           min="0"
           class="form-control"
+          :class="{ 'is-invalid': isInvalid }"
```
safydy-r, line 9: BS `is-invalid` class

src/components/common/ValidationMessage.vue
```diff
@@ -0,0 +1,48 @@
...
+const validationVuexNamespace = namespace('validation')
+
+@Component({
+  components: {
+    BAlert
+  }
+})
+class ValidationMessage extends mixins(Lang) {
+  @Prop() messageKey!: string
+  @Prop() showFixAction?: boolean
```
safydy-r, line 31: Will add the link based on this value

safydy-r: Done

src/components/common/NumericEditor.vue
```diff
@@ -18,10 +19,13 @@
 </template>
 
 <script>
-import { mapMutations } from 'vuex'
-
 export default {
   props: {
+    state: {
```
bzabos, line 24: To make it official: Is this to be the pattern moving forward? All editors will then receive state variable that matches up to bootstrap's API?

I'd like to suggest we rename to validationState for clarity, since the word state is pretty overloaded already!

src/components/common/NumericEditor.vue
```diff
@@ -18,10 +19,13 @@
...
+      type: Boolean,
+      default: undefined, // to tell boostrap `No state`
```
bzabos, line 26: Can this be null instead of undefined?

bzabos (approved):
@safydy-r This is good! I've left a couple minor comments. Overall are you happy with this approach? Any changes you'd like to see now that you've given it a test run?

src/components/common/NumericEditor.vue
safydy-r: @bzabos , yes, we will use that pattern for @poojahpatil90 's ticket to match up to bootstrap's API.
I will rename the prop to `validationState`

src/components/common/NumericEditor.vue
safydy-r: Sure, changing

safydy-r:
Thanks @bzabos for the review.
About this 
> Overall are you happy with this approach? Any changes you'd like to see now that you've given it a test run?

I'm happy with it, we will see how it works with Pooja's validation API tickets. She always pings me in case of missing thing, so I think we're good for now.

safydy-r:
**Note**: for now we don't have validation message yet in storybook side. In fact, stories don't have watchers on activeBlock & activeFlow to run validate_block & validate_flow.

If we want that in the future, an idea would be to move watchers inside a mixin, then reuse it for both `BuilderCanvas.vue` & storybook `BaseMountedVueClass`

yeboaasare (approved):

PR #64, opened 2021-04-14
VMO-3746 [Viamo Blocks] Add #extends slot to Community Flow Builder blocks

poojahpatil90:
Added extras slot above the options section which shows the set first block button. The screenshot shows some test label and input field that I added(but removed in this PR).

image


bzabos (approved):
Looks good to me, thanks!

PR #65, opened 2021-04-16
VMO-3703 Canvas deep linking

poojahpatil90:
Uploading Screen Recording 720p.mov…


src/components/interaction-designer/block-editors/MaxDigitEditor.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="hasIvr" class="form-group block-max-digits">
+  <div v-if="hasIvr" class="form-group block-max-digits" :id="`${block.uuid}.config.ivr.maxDigits`">
```
safydy-r, line 2: Let's be aware of these casing, if vmo-3553 (consume flow-runner v1.0) gets merged before this PR, we need to change the case here and in similar location.

src/components/interaction-designer/block-editors/LabelEditor.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="block-label">
+  <div class="block-label" :id="`${block.uuid}.label`">
```
safydy-r, line 2: Shouldn't we add a `/` after `${block.uuid}` according to AC ?

src/components/interaction-designer/block-editors/MaxDurationSecondsEditor.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="hasIvr" class="form-group block-max-duration-seconds">
+  <div v-if="hasIvr" class="form-group block-max-duration-seconds" :id="`${block.uuid}.config.ivr.maxDurationSeconds`">
```
safydy-r, line 2: I propose we add `:ref="${block.uuid}/..."` too (same value as `:id`), for vuejs DOM manipulation best practice. What do you think @bzabos ?

src/router/index.ts
```diff
@@ -10,9 +10,20 @@ export const routes: Array<RouteConfig> = [
   ...treesRoutes,
 ]
 
-const router = new VueRouter({
+const scrollBehavior = (to) => {
+  if (to.params.field) {
+    const anchor = `${to.params.blockId}.${to.params.field}`
+    const ele = document.getElementById(anchor)
```
safydy-r, line 16: let's avoid non standard or non frequently used abbreviation `ele` here

src/router/index.ts
```diff
@@ -10,9 +10,20 @@ export const routes: Array<RouteConfig> = [
...
+    if (ele) {
+      ele.scrollIntoView({ behavior: 'smooth', block: 'center' })
+    }
```
safydy-r, line 19: can we add an `else` and log a debug here? it would be interesting to see something if the element was not found
eg: `console.debug('Dee linking: cannot scroll to ', anchor, ' as the DOM element is not found')`

src/router/trees.js
```diff
@@ -13,29 +13,40 @@ export const routes = [
...
-  {
-    path: '/trees/create', /* no-op */
-  },
   {
     path: '/trees/:id/interaction-designer/:mode',
+    alias: '/flows/:id/interaction-designer/:mode',
+    name: 'flow-canvas',
     props: (route) => ({ id: route.params.id, mode: route.params.mode }),
     component: () => import(/* webpackChunkName:"chunk-builder" */ '@/views/InteractionDesigner.vue'),
-  }, {
```
safydy-r, line 23: Can you explain me why we deleted these ?

src/store/flow/flow.ts
```diff
@@ -332,7 +333,11 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
     }
 
     commit('flow_addBlock', { block: duplicatedBlock })
-    commit('builder/activateBlock', { blockId: duplicatedBlock.uuid }, { root: true })
```
safydy-r, line 335: Wondering why we deleted this line ? This is telling the store the new duplicated block is the activeBlock, meaning this block will be highlighted in the canvas

src/views/InteractionDesigner.vue
```diff
@@ -208,8 +215,28 @@ export default {
     this.deselectBlocks()
     this.discoverTallestBlockForDesignerWorkspaceHeight({ aboveTallest: true })
 
+    setTimeout(() => {
```
safydy-r, line 218: Did you use `setTimeout` to make sure the `document.querySelector()` finds the element ?
If yes, I suggest we use the vue way with `nextTick`
- similar implementation : https://github.com/FLOIP/flow-builder/pull/33/files#diff-43f7c5f1a1252fa694baadbe4e2a0ec2f48c3a9041968f25ed1cfab7b4bdff9aR166
- official doc: https://vuejs.org/v2/api/#Vue-nextTick

src/views/InteractionDesigner.vue
```diff
@@ -208,8 +215,28 @@ export default {
...
+      const { blockId, field } = this.$route.params
+      if (blockId) {
+        this.activateBlock({ blockId })
+        const blockEle = document.querySelector(`#block\\/${blockId} .plain-draggable`)
```
safydy-r, line 222: same remark with variable name abbreviation

safydy-r:
Overall it's good @poojahpatil90 , I think you had a good opportunity to play with router. Left some questions & comments.

src/store/flow/flow.ts
poojahpatil90: Deleted this line as we want the block to be select due to the route change. When the route changes with a block id that block will be automatically selected (these changes can be seen below in InteractionDesigner.vue file).

src/views/InteractionDesigner.vue
poojahpatil90: `setTimeout()` had to be added to wait for the child components to be loaded. If the child components are not loaded then we are trying to fetch an element before it is even created. The `nextTick` did not work in this scenario.

src/router/index.ts
poojahpatil90: Done

src/router/index.ts
poojahpatil90: Done

src/router/trees.js
poojahpatil90: I have removed the old routes and re-created them using [nested routing](https://router.vuejs.org/guide/essentials/nested-routes.html). We had some hard-coded routes which would get complicated as we add new routes. Hence, I chose to create nested routing which also provides accurate structure.

src/views/InteractionDesigner.vue
poojahpatil90: Done!

src/components/interaction-designer/block-editors/MaxDigitEditor.vue
poojahpatil90: Thanks for pointing this out. I will be consuming the IDs as part of [VMO-3478](https://viamoinc.atlassian.net/browse/VMO-3478). I will keep an eye out for these changes 👍

src/components/interaction-designer/block-editors/LabelEditor.vue
poojahpatil90: As per AC we want the URL to be in `{blockId}/.config.choices[1].audio` format. I did not want to have a mixed set (/ and .) which is why I chose to use . instead of /. I will think more on this while implementing [VMO-3478](https://viamoinc.atlassian.net/browse/VMO-3478). Does that sound good ?

src/components/interaction-designer/block-editors/MaxDurationSecondsEditor.vue
poojahpatil90: We are using the :id to scroll a particular dom element into view. I am not sure how :ref would add an advantage as we would need to access the dom from ref to `scrollIntoView`. 
Ex: 
```
const [el] = this.$refs[`${block.uuid}.config.ivr.maxDurationSeconds`];
if (el) {
  el.scrollIntoView({ behavior: "smooth", block: "center" });
}
```
@safydy-r  Please help me understand what advantages we will have with adding `:ref`

src/views/InteractionDesigner.vue
```diff
@@ -208,8 +215,28 @@ export default {
...
+        if (blockElement) {
+          blockElement.scrollIntoView({ behavior: 'smooth', block: 'center' })
+        }
+      }
+      if (field) {
+        const ele = document.getElementById(`${blockId}.${field}`)
+        if (ele) {
+          ele.scrollIntoView({ behavior: 'smooth', block: 'center' })
+        }
+      }
```
bzabos, line 232: @poojahpatil90 Maybe we could consolidate this `setTimeout()` handler with our scrollBehaviour defined in routes/index.ts

bzabos (approved):
Hi @poojahpatil90 , this looks good. Left one comment on consolidating our scroll behaviour. 

One other thing we discussed was consolidating our dom access into $refs and try to perform that all from InteractionDesigner, but we can defer that to a future task!

src/views/InteractionDesigner.vue
poojahpatil90: Done!

poojahpatil90:
> Hi @poojahpatil90 , this looks good. Left one comment on consolidating our scroll behaviour.
> 
> One other thing we discussed was consolidating our dom access into $refs and try to perform that all from InteractionDesigner, but we can defer that to a future task!

Thanks @bzabos. I will play with $refs as part of VMO-3478 for dom access.

src/components/interaction-designer/block-editors/LabelEditor.vue
safydy-r: Sounds good

src/components/interaction-designer/block-editors/MaxDurationSecondsEditor.vue
safydy-r: @poojahpatil90 , I'm not an expert on `ref` usage yet. So this may give a summary: https://jai.me/blog/vuejs-ref-attribute (or this stackoverflow opinion https://stackoverflow.com/a/64750868/3256489)

src/router/trees.js
safydy-r: thanks for explaining, I just wanted to make sure it has been deleted on purpose

src/store/flow/flow.ts
safydy-r: Ok

safydy-r (approved):
Approving this for now, I replied to your question about `refs` advantage. I you still have some question, pls reach out to me or direct to @bzabos.

yeboaasare (approved):

PR #66, opened 2021-04-19
VMO-3120 - fix flow lifecycle

jcbashdown:
The second argument for router.replace is onload. To use a query param it should be done like this. I'm 100% I tested this multiple times during the flow lifecycle work so I don't really understand how this became broken but this fixes it!


bzabos (approved):
What the heck! 😹 I don't know how this worked either, but the way you've got it now looks right to me !

PR #69, opened 2021-04-20
VMO-3614 + VMO-3615 + VMO-3616 + VMO-3375 - Validation of all block types

poojahpatil90:


src/components/common/FloatEditor.vue
```diff
@@ -51,6 +52,17 @@ export default {
       required: false,
       default: '[0-9\-.,]',
     },
+    validationState: {
```
safydy-r, line 55: I'm realising we repeat these part of code in other component. Wondering if you're happy to move it in a `mixins` way ? Like we did with `src/lib/filters/lang.ts`, we just need to do something similar in each component
```
// if class-based component, eg: components/interaction-designer/Block.vue
import { lang } from '@/lib/filters/lang'

export default {
  components: {
     ...
  }
  mixins: [lang],
  ...
}

// if class-based component, eg: components/interaction-designer/toolbar/TreeBuilderToolbar.vue
import Lang from '@/lib/filters/lang'

@Component({
  components: {
    ...
  },
})
export default class TreeBuilderToolbar extends mixins(Routes, Permissions, Lang) {
  ...
}
```
I can pair with you if you'd like we do this together.

src/components/interaction-designer/block-types/Core_RunFlowBlock.vue
```diff
@@ -4,23 +4,31 @@
...
+      <validation-message :message-key="`block/${block.uuid}/.semantic_label`" #input-control="{ isValid: isSemanticLabelValid }">
+        <block-semantic-label-editor :block="block" :validationState="!isSemanticLabelValid" />
+      </validation-message>
 
-      <div class="form-group">
-        <label>{{ 'flow-builder.destination-flow' | trans }}</label>
+      <validation-message :message-key="`block/${block.uuid}/.config.flow_id`" #input-control="{ isValid: isFlowIdValid }">
+        <div class="form-group">
+          <label>{{ 'flow-builder.destination-flow' | trans }}</label>
+          <select class="form-control" v-model="destinationFlowId" :class="{ 'is-invalid': isFlowIdValid === false }">
```
safydy-r, line 20: `isFlowIdValid` is a boolean, so let's use `!isFlowIdValid` instead of `isFlowIdValid === false`

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
```diff
@@ -24,7 +24,7 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
   async setValidationMaximum({ commit, rootGetters }, { blockId, value }: { blockId: IBlock['uuid']; value: number | string }) {
     commit('flow/block_updateConfigByKey', {
       blockId,
-      key: 'validation_minimum',
+      key: 'validation_maximum',
```
safydy-r, line 27: thanks for this :+1:

src/components/interaction-designer/block-types/ConsoleIO_PrintBlock.vue
```diff
@@ -5,9 +5,15 @@
     </h3>
 
     <fieldset :disabled="!isEditable">
-      <block-name-editor :block="block" />
-      <block-label-editor :block="block" />
-      <block-semantic-label-editor :block="block" />
+      <validation-message :message-key="`block/${block.uuid}/.name`" #input-control="{ isValid: isNameValid }">
+        <block-name-editor :block="block" :validationState="!isNameValid" />
```
safydy-r, line 9: @poojahpatil90 , what if don't add the `<validation-message>` in this level, but under the `<block-....-editor>` component level ? We still have access to `:block` prop there, right ? This avoids us repetitive codes too.

I know you just picked the example I provided from `FlowEditor`, but now I'm wondering if there would be a best & cleaner approach.

poojahpatil90: This would optimize our code and reduce redundant code. Made this change where ever it was possible

src/components/interaction-designer/block-types/Core_RunFlowBlock.vue
poojahpatil90: Interesting thing I noticed when I tried making this change. It does not work with !isFlowIdValid, but works with just isFlowIdValid. I noticed this in other places too!  I guess the value coming in for isValid is actually the value for isInvalid

src/components/common/FloatEditor.vue
poojahpatil90: I am not sure if we still need this change as the code repetition has reduced significantly. Take a look at the new changes and let me know your thoughts.

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
```diff
@@ -5,24 +5,28 @@
...
-        </vue-multiselect>
-      </div>
-
-      <group-selector :block="block"/>
+      <block-name-editor :block="block" />
+      <block-label-editor :block="block" />
+      <block-semantic-label-editor :block="block" />
+
+      <!-- TODO: Verify the type of is_member(boolean or string) and then enable validation for this field -->
+<!--      <validation-message :message-key="`block/${block.uuid}/.config.is_member`" #input-control="{ isValid: isGroupAMemberValid }">-->
```
poojahpatil90, line 13: Commented this line for now until we get clarity on this field. 
@bzabos Can you confirm if we want is_member to be a boolean?
Based on the [flow spec](https://floip.gitbook.io/flow-specification/layers/blocks#set-group-membership) it can be either boolean or an expression 
cc: @safydy @safydy-r

src/components/interaction-designer/block-editors/ContactPropertySelector.vue
```diff
@@ -1,16 +1,19 @@
...
-                     track-by="id"
-                     label="displayLabel"
-                     :placeholder="'flow-builder.contact-property-selector-placeholder' | trans"
-                     :options="subscriberPropertyFields || []"
-                     :allow-empty="false"
-                     :show-labels="false"
-                     :searchable="true">
-    </vue-multiselect>
-  </div>
+  <validation-message :message-key="`block/${block.uuid}/.config.set_contact_property.property_key`" #input-control="{ isValid: isInvalid }">
```
safydy-r, line 2: I think this will be `{ isValid: isInvalid }` after you update from the fix i made in vmo-3704. ;)

src/components/interaction-designer/block-editors/ContactPropertySelector.vue
```diff
@@ -71,3 +76,9 @@ class ContactPropertySelector extends mixins(Lang) {
 
 export default ContactPropertySelector;
 </script>
+
+<style lang="css" scoped>
```
safydy-r, line 80: Let's confirm what @bzabos thinks about it.

Vue-multiselect doesn't use the same approach with Bootstrap for validation. It is using more complicated library https://vuelidate.js.org/

For now, I think this simple css overriding is OK.

safydy-r (commented):
This is good @poojahpatil90 , I will be happy to approve once you push your last change related to the `isInvalid` & you update the storyshot.

src/components/common/FloatEditor.vue
safydy-r: Let's skip for now @poojahpatil90 , we have less repetition than before after you made the 2nd changes. Thanks

src/components/interaction-designer/block-editors/ContactPropertySelector.vue
poojahpatil90: Done!

src/components/interaction-designer/block-editors/ContactPropertySelector.vue
poojahpatil90: This was the simplest solution I could think of instead of installing a new library and configuring it. Also, it does not give the same styling(red box) as bootstrap.

src/components/interaction-designer/block-editors/FormatStringEditor.vue
```diff
@@ -1,18 +1,23 @@
...
-    <text-editor v-model="formatString"
-        :label="'flow-builder.format-string' | trans"
-        :placeholder="'flow-builder.enter-format-string' | trans"/>
-  </div>
+  <validation-message :message-key="`block/${block.uuid}/.config.format_string`" #input-control="{ isInvalid }">
+    <div class="block-format-string">
+      <text-editor v-model="formatString"
+          :label="'flow-builder.format-string' | trans"
+          :placeholder="'flow-builder.enter-format-string' | trans"
+          :validationState="isInvalid" />
```
safydy-r, line 7: @poojahpatil90 , I think we should have the same approach as https://bootstrap-vue.org/docs/components/form-input#contextual-states, because in some component we do use `bootstrap-vue`. The `validationState` prop is false for invalid, true for valid. So this line would be `:validationState="!isInvalid"` and we should update the `validationState` like this https://github.com/FLOIP/flow-builder/pull/69/files#diff-dd8be5f9484a7f9a30bd08dc9d48608cbd1a69b213ba09824f26682a6a42b15eR6 into `:class="{ 'is-invalid': !validationState }"`

What do you think ?

poojahpatil90: Done!

safydy-r (approved):
Thanks for making the changes, approved

src/components/interaction-designer/block-types/Core_SetGroupMembershipBlock.vue
bzabos: Nice one! This one should be an expression editor; if the author would like to hard-code a boolean, I believe they can do so with a static expression! :)

yeboaasare:
@poojahpatil90  I got these errors in the console

yeboaasare:
@poojahpatil90 Can the validation warnings be improved as a user may struggle to understand this?
e.g
Screenshot 2021-05-05 at 12 36 30 PM

poojahpatil90:
@yeboaasare -  We might need to wait for @safydy-r as he is the expert with the [changes](https://github.com/FLOIP/flow-builder/pull/61).

yeboaasare:
@poojahpatil90 @safydy-r  - I encountered this new error in the console. The error is thrown when any block is added to the flow conavas
Screenshot 2021-05-18 at 10 16 16 AM

yeboaasare:
@poojahpatil90 @safydy-r  - This error is thrown when location response block is added to the canvas

poojahpatil90:
> @poojahpatil90 @safydy-r - This error is thrown when location response block is added to the canvas
> Screenshot 2021-05-18 at 10 15 45 AM

@yeboaasare  This bug has been fixed.

poojahpatil90:
> @poojahpatil90 @safydy-r - This error is thrown when location response block is added to the canvas
> Screenshot 2021-05-18 at 10 15 45 AM

@yeboaasare Most schema errors have been fixed by @safydy-r. There are a few related to Select Many block pending, which will be handled as part of this ticket - https://viamoinc.atlassian.net/browse/VMO-4048. 
Please go ahead with testing this PR and let me know if you find any other issue.

poojahpatil90:
> > @poojahpatil90 @safydy-r - This error is thrown when location response block is added to the canvas
> > Screenshot 2021-05-18 at 10 15 45 AM
> 
> @yeboaasare Most schema errors have been fixed by @safydy-r. There are a few related to Select Many block pending, which will be handled as part of this ticket - https://viamoinc.atlassian.net/browse/VMO-4048.
> Please go ahead with testing this PR and let me know if you find any other issue.


@yeboaasare  Updated this PR with fix for schema error on Select Many block as well (fixed by @safydy-r). Ticket VMO-4084 has been resolved. Let me know if you find any other issues.  Thanks in advance!

yeboaasare (approved):

PR #71, opened 2021-04-22
VMO-3576: Error when dragging a block after deleting one expression

poojahpatil90:


safydy-r:
Wow, this is very clean @poojahpatil90 . Thank :100:

safydy-r (approved):

safydy-r:
Pls update from master and check TS errors.
After, you should not have this `node-sass` related build error

src/components/common/PlainDraggable.vue
```diff
@@ -90,6 +90,7 @@ export default {
   },
 
   destroyed() {
+    this.$emit('destroyed')
```
bzabos, line 93: Can you add `{ draggable }` as payload to this event?

bzabos (approved):
This is perfect @poojahpatil90! Thanks! :)

src/components/common/PlainDraggable.vue
poojahpatil90: Sure, I will @bzabos!  It was not required in this context so I hadn't added.

PR #74, opened 2021-04-26
Vmo 3164 output branching config for select1 n

safydy-r:
[UI overview]
- added branching option
- always put `Error` exit at the end
image


src/components/interaction-designer/Block.vue
```diff
@@ -54,7 +54,7 @@
             <span class="badge badge-primary tree-block-item-label tree-block-item-output-subscribers-1"></span>
           </div>
 
-          <h3 class="block-exit-tag badge badge-warning">{{visibleExitTag(key, exit)}}</h3>
+          <h3 class="block-exit-tag badge badge-warning">{{exitTagText(key, exit)}}</h3>
```
safydy-r, line 57: amended this to avoid confusion with the `visible` concept.

src/components/interaction-designer/Block.vue
```diff
@@ -197,6 +197,13 @@ export default {
       return this.block.exits.length
     },
 
+    visibleExits() {
+      return filter(this.block.exits, function(exit) {
+        const isVisible = get(exit.config, 'is_visible') //TODO: change lodash.get with direct prop access when `is_visible` is in IBlockExit interface
```
safydy-r, line 202: @bzabos , my proposal here is to add `is_visible` prop in IBlockExit to easily reflect the spec
> This block can be configured to have a single exit, or a number of exits with possibilities

Or should we put it outside `config`, and update `IBlock` interface instead ?

What do you think?

src/components/interaction-designer/block-editors/BlockOutputBranchingConfig.vue
```diff
@@ -0,0 +1,88 @@
...
+    get isBranchingTypeSegregated() {
+      return this.branchingType === BRANCHING_TYPE_SEGREGATED
+    }
+
+    get isBranchingTypeUnified() {
+      return this.branchingType === BRANCHING_TYPE_UNIFIED
+    }
+
+    mounted() {
+      // @ts-ignore TODO: remove this once IBlock has vendor_metadata key, and confirm this path because FLOIP spec doesn't provide a definition for the output branching persistence
```
safydy-r, line 61: should be tackled here https://viamoinc.atlassian.net/browse/VMO-3882

src/components/interaction-designer/block-editors/BlockOutputBranchingConfig.vue
```diff
@@ -0,0 +1,88 @@
...
+      let value = get(this.block.vendor_metadata, 'branchingType')
+      if (isNil(value) && this.block) {
+        this.block_updateVendorMetadataByPath({ blockId: this.block.uuid, path: 'branchingType', value: this.branchingType })
```
safydy-r, line 64: @bzabos , this is another proposal I made. In fact, we don't have a prop to save the `branchingType` in the floip spec. In voto5, we save it inside `customData.branching`. I didn't hesitate to put it inside vendor_metadata because we already add something like block UI position there.

Any concerns ?

src/components/interaction-designer/block-types/MobilePrimitives_SelectManyResponseBlock.vue
```diff
@@ -63,40 +66,23 @@
...
-import BlockId from '../block-editors/BlockId.vue'
 
 import SelectOneResponseBlock from './MobilePrimitives_SelectOneResponseBlock.vue'
+import {IInflatedChoicesInterface} from "@/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore";
 
 const blockVuexNamespace = namespace(`flow/${BLOCK_TYPE}`)
 const builderVuexNamespace = namespace('builder')
 
-@Component<any>({
-  components: {
```
safydy-r, line 86: As the inheritance seems working fine, I removed these

src/components/interaction-designer/block-types/MobilePrimitives_SelectOneResponseBlock.vue
```diff
@@ -131,10 +138,16 @@ export class MobilePrimitives_SelectOneResponseBlock extends mixins(Lang) {
       this.editEmptyChoice( { choice: oldChoice as IInflatedChoicesInterface })
     }
 
+  handleBranchingTypeChange(isSegregatedBranching: boolean) {
+    this.updateInflatedEmptyChoiceVisibility({ value: isSegregatedBranching })
+  }
+
     @flowVuexNamespace.Getter resourcesByUuid!: {[key: string]: IResourceDefinition}
```
safydy-r, line 145: Will fix these wrong indentations once the PR is approved

bzabos (changes requested):
Hi @safydy! I think we forgot to do a check-in to ensure we were all on the same page for architecture on this one! 😛 

Sad story! I have interpreted the requirements a bit different in terms of how the underlying data will look. From what I understand, `block.exits` will always be our source of truth for what will be presented to the user. 

From that perspective, the action dispatched from selecting either UNIFIED or SEGREGATED would actually rewrite the exits list on the block as necessary to match what would be seen on the canvas. In this case, we wouldn't need to write anything in customData dictionaries.

To detect if we're in one mode or the other, our condition could be something like: compare the list of exits on the block and see if it matches the choices (there are some nuanced bits like: if there's only one choice entered, it looks the same as unified, with the difference being in UNIFIED mode, our first exit will have `default === true`).

Let me know if you'd like to connect on this tomorrow!

src/components/interaction-designer/block-editors/BlockOutputBranchingConfig.vue
safydy-r: After call: we will leave it under vendor_metadata, & prefix it with `io_viamo`

src/components/interaction-designer/Block.vue
safydy-r: After a call with Brett, will avoid using the `is_visible` approach to make sure the block.exits is our source of true for what we have in UI.
Miro note: https://miro.com/app/board/o9J_lXulmu4=/?moveToWidget=3074457356308530338&cot=14

safydy-r: @bzabos : it's no longer using `is_visible` approach according to our last chat.

src/components/interaction-designer/Block.vue
safydy-r: Left the change even we won't have this `is_visible` approach

src/components/interaction-designer/block-types/MobilePrimitives_SelectOneResponseBlock.vue
```diff
@@ -84,8 +88,10 @@ import BlockSemanticLabelEditor from '../block-editors/SemanticLabelEditor.vue'
 import BlockExitSemanticLabelEditor from '../block-editors/ExitSemanticLabelEditor.vue'
 import FirstBlockEditorButton from '../flow-editors/FirstBlockEditorButton.vue'
 import ResourceEditor from '../resource-editors/ResourceEditor.vue'
+import BlockOutputBranchingConfig from '../block-editors/BlockOutputBranchingConfig.vue'
 import BlockId from '../block-editors/BlockId.vue'
 import { mixins } from 'vue-class-component'
+import { set } from 'lodash'
```
safydy-r: will clean this unused import

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
```diff
@@ -44,19 +44,14 @@ export const getters: GetterTree<ICustomFlowState, IRootState> = {
...
+      const exit = findExitFromResourceUuid(resourceUuid, currentBlock, !getters.isExitsBranchingSegregated)
       memo[choiceKey] = {
-        exit: getters.blockExitFromResourceUuid(resourceUuid),
+        exit,
         resource: rootGetters['flow/resourcesByUuid'][resourceUuid]
       }
       return memo
     }, choices)
   },
-  blockExitFromResourceUuid: (state, getters, rootState, rootGetters) => (resourceUuid: string): IBlockExit => {
```
safydy-r, line 54: took the opportunity to remove dynamic getter in favor of `findExitFromResourceUuid` function

safydy-r:
@bzabos , this PR is ready for review again after I implemented what we have discussed

src/store/flow/block.ts
```diff
@@ -49,13 +46,31 @@ export const mutations: MutationTree<IFlowsState> = {
...
+    if (!isNil(insertAtIndex)) { // insertion, eg: case for block types having branching exits option (Segregated | Unified)
+      blockExits.splice(insertAtIndex, 0, newExit);
+    } else { // just push
+      blockExits.push(newExit);
+    }
+  },
+  block_updateExits(state, { block, newExits, shouldUseCache = false }: { block: IBlock; newExits: IBlockExit[]; shouldUseCache: boolean }) {
+    if (shouldUseCache) {
+      // @ts-ignore: TODO: remove this once IBlock has vendor_metadata key
+      set(block.vendor_metadata, 'io_viamo.cache.outputBranching.segregatedExits', newExits);
```
safydy-r, line 70: [] @bzabos will take time to look at the PR
[] @safydy-r  will re-confirm with Jon if select many really needs segregated exits

bzabos: @safydy For future reference: via `voto5` this feature is used in:
image

safydy-r: Jon confirmed in Jira ticket, that always using `Unified exits` for select many would make the most sense

safydy-r: @bzabos , about the voto5 reference, I already copied this component from voto5 and adapted it for flow-build (store, props, etc).

PR #86, opened 2021-05-19
Vmo 3616 correct missed schema

safydy-r:
I fixed most the schema issue.

But the selectMany still has issues, I don't know why ? 

I created this ticket to handle the fix: https://viamoinc.atlassian.net/browse/VMO-4048


poojahpatil90 (approved):
This is great!  Thanks @safydy-r 🥇

PR #92, opened 2021-05-28
Vmo 3942 allow block multiselection

safydy-r:
[Screenshots]
Screen Shot 2021-05-27 at 9 52 21 PM


src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -1,150 +1,155 @@
 <template>
   <div class="tree-builder-toolbar">
```
safydy-r, line 2: The diff seems hard to follow here, but what I did in this template was very simple:
- renamed the class `tree-builder-toolbar` into `tree-builder-toolbar-main-menu`
- created a new div wrapper, and named it `tree-builder-toolbar`
- put the new alert banner within this new added div `tree-builder-toolbar` (check here: https://github.com/FLOIP/flow-builder/pull/92/files#diff-3db5c56652da6d4f7b33c23e432a38b9b73d0b30d47df648929798e4c82d498bR147)

src/store/builder/index.ts
```diff
@@ -1,5 +1,5 @@
 import {
-  cloneDeep, flatMap, isEqual, keyBy, map, mapValues, get, filter, union,
+  cloneDeep, flatMap, isEqual, keyBy, map, mapValues, get, filter, union, includes
```
safydy-r, line 2: Will remove this unused import

src/components/interaction-designer/Block.vue
```diff
@@ -250,6 +264,10 @@ export default {
       return !!filter(activeConnectionsContext, (context) => context.sourceId === block.uuid || context.targetId === block.uuid).length
     },
 
+    isBlockSelected() {
+      return this.block.vendor_metadata.io_viamo.uiData.isSelected
```
jcbashdown, line 268: This seems weird... Now I think about it, it's not really viamo specific to even store the block position in io_viamo as it's an integral part of the builder

I suppose this isn't something we want to think too deeply about right now though. It also seems a bit weird to me that if you export and import the flow it could already have blocks selected but it's not necessarily a bad thing... Happy to leave this right now.

jcbashdown (changes requested):
Looks good. Will approve once the conflicts are resolved and that unused lodash import is removed.

I have this concern about storing the selection status in platform metadata. It seems weird to me vs just having some list of uuids we check against - e.g. using selectedBlocks. I don't think I feel so strongly that I need this change to approve but I'm interested in what you think?

src/components/interaction-designer/Block.vue
```diff
@@ -12,11 +12,25 @@
...
-        @dragStarted="selectBlock">
+        @dragStarted="activateBlock({ blockId: block.uuid })">
 
       <div class="d-flex justify-content-between">
         <div class="header-actions-left">
-<!--          <font-awesome-icon :icon="['fas', 'check-circle']" class="fa-btn" :style="{color: '#0069d9'}"/>-->
+          <!--Selection-->
+          <font-awesome-icon
+            v-if="isBlockSelected"
+            :icon="['fas', 'check-circle']"
```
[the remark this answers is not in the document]
safydy-r: hihi, yeah ... I just followed the provided design ;)

src/components/interaction-designer/Block.vue
safydy-r: @jcbashdown , I agree on the fact that block positions are not `io_viamo`, maybe we decided to do so because viamo blocks persist position using this field id?

About new introcuded field, let me see if I can move it out of `io_viamo`.

safydy-r:
@jcbashdown , about this one

> I have this concern about storing the selection status in platform metadata. It seems weird to me vs just having some list of uuids we check against - e.g. using selectedBlocks. I don't think I feel so strongly that I need this change to approve but I'm interested in what you think?

I think you're right here, no need for now to persist the `isSelected` state. I will remove it and use only a simple list of uuids for the implementation. Thanks

jcbashdown (approved):

jcbashdown:
Looks good! I like this improvement.

safydy-r:
Note for me (or the one who will merge this): pay attention to the fact that vmo-3478 https://github.com/FLOIP/flow-builder/pull/88#pullrequestreview-675786060 might be merged before. If that's the case, we may need small adjustment to this PR.

cc: @poojahpatil90

yeboaasare (approved):

PR #93, opened 2021-05-28
VMO-690 - Fixes for updated flow spec

jcbashdown:


safydy-r (approved):
I think it's good @jcbashdown  ,
Thanks you for fixing eslints as well ;)

PR #94, opened 2021-05-28
Vmo 3943 allow multiple deletion and duplicate

safydy-r:
Just to note that multiselection part is done in vmo-3942

[screenshots]
After selecting blocks
Screen Shot 2021-05-28 at 10 56 29 AM

Before deleting selected blocks
Screen Shot 2021-05-28 at 10 56 42 AM


src/components/interaction-designer/toolbar/SelectionBanner.vue
```diff
@@ -2,11 +2,44 @@
...
+                @click="deleting = false">
+          {{ 'flow-builder.no-cancel' | trans }}
+        </button>
+
+        <button class="btn btn-danger btn-sm mr-2"
+                @click="confirmMultipleDeletion">
+          {{ 'flow-builder.yes-delete' | trans }}
+        </button>
       </div>
-      <button class="btn btn-outline-dark btn-sm" @click="block_clearMultiSelection">
```
safydy-r, line 7: moved inside `<!--Main multiselect actions-->` part

poojahpatil90 (approved):
This looks good overall!  I am curious on what would happen if a user click on the trash icon on a block when multiple blocks are selected?

safydy-r:
@poojahpatil90 , thanks for reviewing it ;).
And good spot for this one
> I am curious on what would happen if a user click on the trash icon on a block when multiple blocks are selected?
I updated the delete action handler to remove the block from the `selectedBlocks` list before we delete it. That way we have everything update.

And for this
> Also, I reduced the banner height for VMO-3478. We might need to come up with a consistent style for the banner. Let me know your thoughts.
Let's stay in touch. Depending on which ticket gets merged first (vmo-3943 or vmo-3478) we may need to update the last PR.

yeboaasare:
@safydy-r  when there are many blocks on a canvas, those at the top tend to overlap the banner. Users have to scroll or move blocks to other parts of the canvas before they can click on the "duplicate" or "delete" buttons

Screenshot 2021-06-04 at 12 05 39 PM

yeboaasare:
@safydy-r  - I also observed that selected blocks lose their selection after they have been duplicated.
A flow user would have to select the blocks again if they plan to make another duplication.

safydy-r:
@yeboaasare , Thanks for catching these 👍 
I made few changes:
- Always put the banner on top
- keep duplicated blocks selected
You can retest again

yeboaasare:
@safydy-r  - I get this console log when I drag a duplicated block(i.e Read,Print,Log,Run , set Group membership,Location,Photo blocks). This only happens when the block was first selected and duplicated. It doesn't happen when the block is duplicated using the duplicate icon inside the block
Screenshot 2021-06-07 at 9 55 01 AM

safydy-r:
@yeboaasare , I pushed a fix. You may have some `warn` console 
> Block onMoved positioning draggable on e0d5e8b0-e479-4f65-86e9-d07fb22f495a can't access property "initElm", props is undefined
but it's ok for now, as long as the block could be duplicated, could be moved correctly, could be deleted.

Thanks 🙏

yeboaasare (approved):

PR #95, opened 2021-06-01
VMO-3865   Deep linking - rename routes

poojahpatil90:
Note:  Renaming routes to have consistent new standard of having routes as /flows instead of /trees


builder.config.json
```diff
@@ -39,15 +39,15 @@
...
           "name": "editTree",
-          "path": "\/trees\/{treeId}\/{component?}\/{mode?}",
-          "params": ["treeId", "component", "mode"],
+          "path": "\/flows\/{flowId}\/{component?}\/{mode?}",
+          "params": ["flowId", "component", "mode"],
           "methods": ["GET", "HEAD"]
-        },
+        }
+      },
+      "trees": {
```
safydy-r, line 50: @poojahpatil90 , Wondering why we moved out only `editTree` to `"flows"` not the whole thing under `"trees"` ?

safydy-r (changes requested):
Thanks for making naming consistent.
I have few concerns:
1- how about this one https://github.com/FLOIP/flow-builder/blob/master/src/views/InteractionDesigner.vue#L311 ?
2- can you pls update in Readme.md to make sure mentioned routes are working ? (those with `trees/`). 
I think we need to change them to / (home) and /flows/new (newFlow)
3- I know we don't use resource viewer yet, but I'd suggest we update routes there as well /src/views/ResourceViewer.vue. Do you agree?
4- don't forget to update storyshot as snapshots have routes references too

builder.config.json
poojahpatil90: @safydy-r  Good question!  Currently we have `editTree` route being actively used. I believe the other routes like `resumeableAudioUpload` etc have not yet been used in flow builder. I am concerned that moving the other routes might break something in the future when we start using them. It would be better if we move it when we use the other routes as we can test that it does not break things. 
Let me know your thoughts on this?

poojahpatil90:
> Thanks for making naming consistent.
> I have few concerns:
> 1- how about this one https://github.com/FLOIP/flow-builder/blob/master/src/views/InteractionDesigner.vue#L311 ?
> 2- can you pls update in Readme.md to make sure mentioned routes are working ? (those with `trees/`).
> I think we need to change them to / (home) and /flows/new (newFlow)
> 3- I know we don't use resource viewer yet, but I'd suggest we update routes there as well /src/views/ResourceViewer.vue. Do you agree?
> 4- don't forget to update storyshot as snapshots have routes references too

1. I am sorry, I do not get what you were trying to say with point 1. Can you explain a little bit about it?
2. Good point!  I updated Readme.md, please review it once
3. I feel it would be better to update it when we enable and work on ResourceViewer.vue. The reason being, I cannot test any changes that I make now and things will not work unless the route is updated anyway. Can I leave it to the future? I can create a ticket or update the ticket on Resource Viewer with this info?
4. Done!

builder.config.json
safydy-r: Thanks for clarifying.
I'm still hesitating because we cannot leave a TODO here to remind us in the future.
But, let's leave it as it is for now then.

safydy-r (approved):

builder.config.json
```diff
@@ -39,15 +39,15 @@
           "name": "home",
           "path": "\/",
           "methods": ["GET"]
-        }
-      },
-      "trees": {
+        },
         "editTree": {
```
jcbashdown, line 43: @poojahpatil90 I think this will require some changes in voto5 - see the `app/Http/Controllers/FlowInterop/FlowsController.php` index method and `resources/assets/js/floip/builder/views/CustomInteractionDesigner.vue`

Also, how come this is still editTrees rather than editFlows?

jcbashdown (changes requested):

builder.config.json
jcbashdown: I think there are also some bits of deep linking config missing from the voto5 flow builder router in `resources/assets/js/floip/builder/routes/index.js`

builder.config.json
```diff
@@ -39,15 +39,15 @@
...
           "name": "editTree",
-          "path": "\/trees\/{treeId}\/{component?}\/{mode?}",
-          "params": ["treeId", "component", "mode"],
+          "path": "\/flows\/{flowId}\/{component?}\/{mode?}",
```
jcbashdown, line 45: Do you still need `{component?}` here given this is now static in the router? I might be misunderstanding how this fits together.

builder.config.json
poojahpatil90: I remember keeping it editFlow and not sure how that changed. I fixed the `editTree` to `editFlow` now.

builder.config.json
poojahpatil90: I am guessing we need the `{component?}` to switch between flow and resource views. Do you think we might need something new instead of changing the component?

safydy-r (approved):

builder.config.json
jcbashdown: Okay, so this intended for a future feature? I think this is fine to leave for now.

PR #96, opened 2021-06-03
VMO-4088 - update the router patch

jcbashdown:


scripts/prepare-gh-pages-deployment.sh
```diff
@@ -15,6 +15,7 @@ abort()
 *** - Regenerate patches:
 ***     $ git diff src/router/index.ts > patches/gh-pages-prepare-router.patch
 ***     $ git diff vue.config.js > patches/gh-pages-prepare-vue-config.patch
+***     $ git diff builder.config.js > patches/remove-backend-routes-from-config.patch
```
safydy-r, line 18: I thought `patches/remove-backend-routes-from-config.patch` wasn't tracked yet, but after checking ... it's already in `master`. Thanks

safydy-r (approved):
Thanks for this quick fix

PR #99, opened 2021-06-09
Test build bundle

jcbashdown:
This branch demonstrates the issue with including the standalone router in the dist bundle. The important files are `src/store/flow/flow.ts` and `src/views/InteractionDesigner.vue`

As in VMO-2573 on bitbucket - https://bitbucket.org/voto/voto5/branch/VMO-2573-builder-reintegration - the intention when using the standalone builder as an embedded frontend within another app is that it should be possible to hook into an existing vue-router set up. This means that we should only be referencing the standalone router in `src/main.ts` and no where else as this is not included when you `build-bundle`. If you don't do this you have an error when embedding the result of `build-bundle` -`Uncaught TypeError: Cannot redefine property: $router`

I think the first step to resolve this is to extract `scrollBlockIntoView` and `scrollBehavior` out of `router/index.ts` as it doesn't look like they have direct dependencies in that file. Where is a good place for this?

The second step is trickier, so far as far as I know we have done all the operations like `router.replace()` in components - e.g. https://github.com/FLOIP/flow-builder/blob/master/src/views/ImportFlow.vue#L210

Can we arrange things to that this is available in the stores also such that the router comes from `resources/assets/js/app.js` in voto5 and from `src/main.ts` when running flow-builder as a standalone app?

Once this is done we should let #proj-floip know that the embedded builder is fixed and they can branch from master and `build-bundle` to get a branch which is installable using the git method.


src/store/flow/flow.ts
```diff
@@ -336,10 +336,10 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
 
     commit('flow_addBlock', { block: duplicatedBlock })
 
-    router.replace({
-      name: 'block-selected-details',
-      params: { blockId: duplicatedBlock.uuid },
-    })
+    //router.replace({
```
safydy-r, line 339: @poojahpatil90 , would the deep linking work without these ?

jcbashdown: I assume it wouldn't? Just trying to highlight where the problem is. We'll need a second PR for an actual fix I think?

PR #100, opened 2021-06-10
Vmo 3951 hook import into basic validations

jcbashdown:


src/components/interaction-designer/flow-editors/import/ErrorHandler.vue
```diff
@@ -7,7 +7,7 @@
       </div>
```
safydy-r, line 7: I know the above line was in a previous PR, but I wondering if you can take the opportunity to localize this missed text.

For info: in case of TS error related to `trans()`, check [here](https://github.com/FLOIP/flow-builder/blob/master/src/components/interaction-designer/block-types/ConsoleIO_PrintBlock.vue#L58) how I load the `Lang` mixin within class based component, pay attention to the importation `import Lang from '@/lib/filters/lang'` instead of ` { lang } `

src/lib/filters/lang.ts
```diff
@@ -19,8 +19,8 @@ const createProxy = (name: string) => function () {
   },
 })
 class Lang extends Vue {
-  trans(translation: string) {
-    return thisGlobal.Lang.trans(translation)
+  trans(translation: string, interpolations?: object) {
```
safydy-r, line 22: Good idea here

builder.config.json
```diff
@@ -72,12 +72,12 @@
       {
         "id": "22",
         "label": "English",
-        "iso_639_3": "EN"
+        "iso_639_3": "eng"
```
safydy-r, line 75: Can we also update these pls:
- https://github.com/FLOIP/flow-builder/blob/VMO-3951-hook-import-into-basic-validations/src/store/builder/2019-10-10-shortcut-flow.json#L18
- https://github.com/FLOIP/flow-builder/blob/VMO-3951-hook-import-into-basic-validations/stories/Flow.stories.ts#L57

src/store/builder/5b8c87d6-de90-4bc4-8668-4f040000006e-flow-block-group.json
```diff
@@ -61,9 +67,9 @@
             }
           },
           "type": "Core.SetGroupMembership",
-          "name": "",
+          "name": "def",
           "label": "",
-          "semantic_abel": "",
+          "semantic_label": "",
```
safydy-r, line 72: nice catch :)

src/store/validation/index.ts
```diff
@@ -140,8 +160,11 @@ export function createDefaultJsonSchemaValidatorFactoryFor(jsonSchema: JSONSchem
   if (subSchema === '') {
     return ajv.compile(jsonSchema)
   } else {
-    ajv.addSchema(jsonSchema)
-    const validate = ajv.getSchema(subSchema)
+    let validate = ajv.getSchema(subSchema)
+    if(!validate) {
```
safydy-r, line 164: @jcbashdown , does this invalid scenario occur because the `'#/definitions/IContainer'` is not considered as subSchema for `flowSpecJsonSchema.json` ?
from this code: `validators.set(validationType, createDefaultJsonSchemaValidatorFactoryFor(flowContainerJsonSchema, '#/definitions/IContainer'))`

If yes, what if we do not provide a subSchema like we did with block https://github.com/FLOIP/flow-builder/blob/a6da072d6f61e7be654ba9961f4c5fda616f1f12/src/store/validation/index.ts#L127 ?
If this will work, I suggest we remove this new condition.

safydy-r (commented):
Overall I'm ok, just left few comments/questions

src/store/validation/index.ts
jcbashdown: This avoids an error on the next page `Uncaught (in promise) Error: schema with key or id "" already exists`. I'm not sure it's ideal but I'm struggling to find a better way to do it... Your suggested solution still has this error. 

I think basically it's caused by calling `createDefaultJsonSchemaValidatorFactoryFor` twice with the same json schema - once for individual flows and once for the container. The only thing we do to prevent this is to call `validators.has(validationType)` but this doesn't record that all the validators in https://github.com/FLOIP/flow-runner/blob/master/dist/resources/flowSpecJsonSchema.json are already registered, only that `flow` is. That means that as long as subSchema is not a blank string we add the jsonSchema again each time we hit `ajv.addSchema(jsonSchema)` with the stuff in that file. By checking if there is something already registered for the subschema, we prevent adding the same json twice (as necessarily the sub schema would have been added already if all that json was added).

I don't think doing the check here - as close to the place we add the potentially duplicate schema as possible - is terrible but it does make me wonder about the point of `validators.has()`

builder.config.json
safydy-r: Thanks for changing them

src/components/interaction-designer/flow-editors/import/ErrorHandler.vue
safydy-r: Thanks for making changes here as well

src/store/validation/index.ts
safydy-r: I see, thanks for raising this scenario
> calling createDefaultJsonSchemaValidatorFactoryFor twice with the same json schema
I will keep that in mind

safydy-r (approved):
Thanks @jcbashdown , approving

src/store/validation/index.ts
jcbashdown: Very weird. When I tested this before, removing subSchema from the call didn't fix it but just been debugging another issue and making this change fixes that AND the original issue doesn't come back. Suspecting that I didn't refresh the browser the first time or something dumb like that!

PR #101, opened 2021-06-15
Vmo 3944 allow moving multiple blocks

safydy-r:


src/store/builder/index.ts
```diff
@@ -113,27 +117,113 @@ export const mutations: MutationTree<IBuilderState> = {
     operations[operation.kind] = operation
   },
 
-  setBlockPositionTo(state, { position: { x, y }, block }) {
```
safydy-r, line 116: moved it as actions

src/store/builder/index.ts
```diff
@@ -113,27 +117,113 @@ export const mutations: MutationTree<IBuilderState> = {
...
+            y: block.vendor_metadata.io_viamo.uiData.yPosition,
+          }
+        })
+      }
+
+      // Translate other selected blocks
+      const otherSelectedBlocks = filter(rootGetters['flow/selectedBlocks'], currentBlock => currentBlock.uuid !== block.uuid);
+      forEach(otherSelectedBlocks, async (currentBlock: IBlock) => {
+        let newPosition = {} as IPosition
+        await dispatch('setBlockPositionFromDelta', { delta: translationDelta, block: currentBlock }).then(
```
bzabos, line 195: Hi @safydy-r , a couple things on this one: I am feeling confused by the need to manage block and draggable separately. Can we wrap these two mutations into a single action (eg. `setBlockAndSyncDraggablePositionTo({pos})` or `setBlockAndSyncDraggablePositionFromDelta({delta})`) then just call that everywhere? I think we want them to be the same in most circumstances, right?

src/store/builder/index.ts
```diff
@@ -113,27 +117,113 @@ export const mutations: MutationTree<IBuilderState> = {
...
+
+      // Prepare translation
+      const translationDelta: IPosition = {
+        x: x - initialPosition.x,
+        y: y - initialPosition.y
+      }
+
+      let shouldReversePosition = false;
+      if (translationDelta.x > 0 /* moving to the right */ ||
+        rootGetters['flow/selectedBlockAtTheFurthestLeftPosition'].vendor_metadata.io_viamo.uiData.xPosition > 0 /* left space still available */) {
```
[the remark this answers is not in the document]
bzabos: Since the new getters like `selectedBlockAtTheFurthestLeftPosition` are only used here, can we move them into this file as a modue-level functions that we call directly? This would mitigate iterating every block on the canvas on every change if we're only ever using here :)

src/store/builder/index.ts
```diff
@@ -113,27 +117,113 @@ export const mutations: MutationTree<IBuilderState> = {
...
+        block.vendor_metadata.io_viamo.uiData.xPosition = x
+      } else {
+        translationDelta.x = 0
+        block.vendor_metadata.io_viamo.uiData.xPosition = initialPosition.x
+        shouldReversePosition = true
+      }
+
+      if(translationDelta.y > 0 /* moving to the top */ ||
+        rootGetters['flow/selectedBlockAtTheTopPosition'].vendor_metadata.io_viamo.uiData.yPosition - state.toolbarHeight > 0 /* top space still available */) {
```
bzabos, line 172: This might get hard to maintain as the UI becomes more fancy. It feels fragile to try to match css and logic. Is there a way we can remove the need to track anything to do with toolbar height? My first thought is to use CSS to manage margins:

```
------------------------ Toolbar
------------------------ Alerts Toolbar
|----------------------| Canvas
|----------------------|
|----------------------|
|----------------------|
```

Using the same logic as we already use to show/hide alerts-toolbar, we can (create, then) apply `.has-alerts` class to canvas which can add a margin-top of whatever the magic number is. This mitigates mixing javascript and css, keeps pixel variations within css, and avoids having those magic numbers sprinkled throughout our formulas and vuex state. 

If we use margins instead of padding, then the top of what we can see on the canvas becomes the _true_ top of the canvas and the formulas needn't be concerned. 

What do you think about this?

bzabos (changes requested):
Hi @safydy ! I've left a couple notes on separation of concerns and hopefully reduce a bit of complexity. I'd love a demo sometime, let me know if you have any Qs. :)

src/store/builder/index.ts
```diff
@@ -91,6 +95,26 @@ export const getters: GetterTree<IBuilderState, IRootState> = {
...
+  selectedBlockAtTheFurthestLeftPosition(_state, getters) {
+    return minBy(getters.selectedBlocks, 'vendor_metadata.io_viamo.uiData.xPosition')
+  },
+
+  isAnyLeftSpaceAvailable(_state, getters) {
+    return getters.selectedBlockAtTheFurthestLeftPosition.vendor_metadata.io_viamo.uiData.xPosition > 0
+  },
+
+  isAnyTopSpaceAvailable(state, getters) {
+    return getters.selectedBlockAtTheTopPosition.vendor_metadata.io_viamo.uiData.yPosition - state.toolbarHeight > 0
```
safydy-r, line 116: @bzabos , [you comment about separating pixel](https://github.com/FLOIP/flow-builder/pull/101#discussion_r655485434) computation is still relevant here.
I introduced this `toolbarHeight` because we would like to know in JS side if there is no more space between the topBlock and the toolbar, and we know the toolbar has a dynamic height.

I tried you proposal 
> Using the same logic as we already use to show/hide alerts-toolbar, we can (create, then) apply .has-notification-alerts-banner class to canvas which can add a margin-top of whatever the magic number is. This mitigates mixing javascript and css, keeps pixel variations within css, and avoids having those magic numbers sprinkled throughout our formulas and vuex state.

... but I cannot resolve the challenge of removing this formula from JS. Any idea ?

Screen Shot 2021-06-21 at 3 08 05 PM

PR #103, opened 2021-06-15
Vmo 4175 fix jest encountered an unexpected token

safydy-r:


poojahpatil90 (approved):
@safydy-r  Thanks for fixing this

PR #104, opened 2021-06-23
Vmo 4142 unified blocks set contact property

safydy-r:
- added ContactPropertyEditor
- removed the ContactPropertySelector as we don't need it anymore
- updated storybook

[Screentshots]
New UI
Screen Shot 2021-06-23 at 12 44 35 PM

On storybook:
- updated the `existing data` story
- add new story named `clear action` to reflect the state when we set the action to `Clear contact prop value`
Screen Shot 2021-06-23 at 12 43 32 PM


tests/unit/__snapshots__/storybook.spec.ts.snap
```diff
@@ -82,6 +82,12 @@ exports[`Storyshots Core/Run Another Flow Block Styled Default 1`] = `
 </div>
 `;
 
+exports[`Storyshots Core/Set Contact Property Clear Action 1`] = `
```
jcbashdown, line 85: Obviously not ideal to have the empty snapshot. Is this something you can fix? I know we still have it for other sidebars.

stories/SetContactPropertyBlock.stories.ts
```diff
@@ -50,33 +52,36 @@ export const Default = () => (DefaultClass)
...
-        name: 'comment',
-        displayLabel: 'Comment',
-      },
-    })
+    this.setDescription(blockId)
+    this.block_updateConfigByPath({blockId, path: 'set_contact_property.property_key', value: 'gender'})
+    this.block_updateConfigByPath({blockId, path: 'set_contact_property.property_value', value: 'M'})
   }
 
   @Mutation addSubscriberPropertyField!: ({property}: { property: any }) => void
```
safydy-r, line 63: We don't need this anymore, will remove it and the one in below story before merge

src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
```diff
@@ -0,0 +1,143 @@
...
+            {{ 'flow-builder.clear-contact-property' | trans }}
+          </label>
+        </div>
+      </div>
+
+      <validation-message
+        #input-control="{ isValid }"
+        :message-key="`block/${block.uuid}/config/set_contact_property/property_key`">
+        <div class="block-contact-property-key">
+          <text-editor
```
jcbashdown, line 41: How come we no longer allow selection from builder.config.json properties? This may break the import validations - this should be tested - as we check there that block properties exist in the org (e.g. are loaded in the builder.config.json)

src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
```diff
@@ -0,0 +1,143 @@
...
+      blockId: this.block.uuid,
+      path: 'set_contact_property.property_value',
+      value,
+    })
+  }
+
+  @flowVuexNamespace.Mutation block_updateConfigByPath!: (
+    {blockId, path, value}: { blockId: string, path: string, value: string }
+  ) => void
+  @flowVuexNamespace.Mutation block_updateVendorMetadataByPath!: (
```
jcbashdown, line 137: Is this used anywhere? I understood we were going to set the state of the UI - which radio button is selected - in vendor_metadata. I'm not seeing this anywhere though?

tests/unit/__snapshots__/storybook.spec.ts.snap
safydy-r: I agree with you, this is something we need to fix later, maybe after July delivery

src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
safydy-r: Good catch here, it's unused, I'm removing it

src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
safydy-r: @jcbashdown , Jon & Amen confirmed that last week to change the dropdown selector into an open input field. If I'm not wrong Jon had a discussion with Mark about it.
To my understanding, we want this change for community version which should have allow users to enter any `contact prop` (the dropdown doesn't allow user to add contact prop for now). But in the future, we would probably go back to the dropdown. Need to confirm that with Jon though, because Brett asked the same question during retro meeting

src/store/flow/block.ts
```diff
@@ -59,11 +60,23 @@ export const mutations: MutationTree<IFlowsState> = {
...
+  block_updateConfigByPath(state, {blockId, path, value}: {blockId: string, path: string, value: object | string}) {
+    // Make nested assignment reactive
+    setWith(
+      findBlockOnActiveFlowWith(blockId, state as unknown as IContext).config!,
+      path,
+      value,
+      (nestedValue, key, nestedObject) => Vue.set(nestedObject, key, nestedValue),
+    )
+  },
+  block_updateVendorMetadataByPath(state, {blockId, path, value}: {blockId: string, path: string, value: object | string}) {
```
poojahpatil90, line 72: Just wondering if this mutation is being used anywhere?   I see that you removed it from ContactPropertyEditor

poojahpatil90 (approved):
Overall looks good!  I just left a few questions

poojahpatil90:
> * added ContactPropertyEditor
> * removed the ContactPropertySelector as we don't need it anymore
> * updated storybook
> 
> [Screentshots]
> New UI
> Screen Shot 2021-06-23 at 12 44 35 PM
> 
> On storybook:
> 
> * updated the `existing data` story
> * add new story named `clear action` to reflect the state when we set the action to `Clear contact prop value`
> 
> Screen Shot 2021-06-23 at 12 43 32 PM

@safydy-r Does it make sense to provide some kind of hint that @(null) is equivalent of clearing the contact property value? Or do we expect the user to not have that knowledge?

src/store/flow/block.ts
safydy-r: thanks, I removed this too

PR #107, opened 2021-06-29
Vmo 3895 port contact properties config into all blocks

safydy-r:
[For devs]
- fontawesome issue on storybook should be fixed in https://github.com/FLOIP/flow-builder/pull/119/files#diff-3121640cbdac49238843964b108b1700e17ca2d577e11d7a5e7471b939f86674R33

[For all]
Concerned blocks: all blocks apart from SetContactProperty which is special.
Let's ignore TS errors during `serve` for this PR, they came from the target branch

[Screenshots]
For all blocks, when checkbox not set
Screen Shot 2021-07-19 at 3 49 11 PM

For non interactive blocks, when checkbox is set
Screen Shot 2021-07-19 at 3 49 26 PM

For interactive blocks (with response)
Screen Shot 2021-07-19 at 3 57 16 PM


poojahpatil90:
@safydy-r  It is looking as per Design. One minor thing, can you add a divider(<hr>) between the title and contact property sections?
I would also recommend adding @asalami in the comments for design review.

src/components/interaction-designer/block-editors/GenericContactPropertyEditor.vue
```diff
@@ -0,0 +1,205 @@
+<template>
```
jcbashdown, line 1: I don't really understand why we need so much new code here given how similar the config we are writing is to the existing Contact Property Block config. Can you explain? 

I think I'm just going to go ahead and merge because it seems to be working fine but I'd like to create a ticket to find a way to share this functionality.

jcbashdown (approved):

src/components/interaction-designer/block-editors/GenericContactPropertyEditor.vue
safydy-r: I was trying to reuse code from `components/interaction-designer/block-editors/ContactPropertyEditor.vue`, and realized the component became too much complicated, in fact we have a bunch of conditional rendering which are different for SetContactPropBlock and the other Block.

So I'd like to keep it separated.

PR #109, opened 2021-07-08
VMO-4138 Unified Message Block

poojahpatil90:
Note: I need to add a few icons in the resources part. Waiting for the design team to provide the icons as it was not downloadable from the Figma link provided.

https://user-images.githubusercontent.com/23651142/124847563-fde53700-df4f-11eb-8815-cca751afe4c9.mov

Update: Added icons and dividers

https://user-images.githubusercontent.com/23651142/125005316-4ebc6480-e010-11eb-89dc-6e958953ceb2.mov

Update: After fixing review comments + adding block semantic label

Screen Shot 2021-07-11 at 1 11 14 PM


src/components/interaction-designer/resource-editors/ResourceEditor.vue
```diff
@@ -1,92 +1,89 @@
...
+              <div v-if="contentType === SupportedContentType.AUDIO">
+                <audio-library-selector
+                  :audio-files="availableAudio"
+                  :lang-id="languageId"
+                  :resource-id="resource.uuid"
+                  :selected-audio-file="findOrGenerateStubbedVariantOn(
+                    resource,
+                    {language_id: languageId, content_type: contentType, modes: [mode]}).value" />
+
+                <phone-recorder
```
bzabos, line 47: I think the mockup I was looking at had Upload Audio button first, then Record Audio second. Is that what you're seeing as well? Should we swap them?

src/components/interaction-designer/resource-editors/ResourceEditor.vue
```diff
@@ -159,6 +161,16 @@ export class ResourceEditor extends mixins(FlowUploader, Permissions, Routes, La
...
   SupportedContentType = SupportedContentType
 
+  iconsMap = new Map<string, string>()
+
+  created() {
+    this.iconsMap.set(SupportedMode.SMS, 'message')
+    this.iconsMap.set(SupportedMode.USSD, 'ussd')
+    this.iconsMap.set(SupportedMode.IVR, 'audio')
+    this.iconsMap.set(SupportedMode.RICH_MESSAGING, 'rich-messaging')
+    this.iconsMap.set(SupportedMode.OFFLINE, 'phone')
```
bzabos, line 171: FYI, we could initialize iconsMap with its values if you want to couple these directly to a single reference site: 
```
iconsMap = new Map<string, string>([
  [SupportedMode.SMS, 'message'],
  [SupportedMode.USSD, 'ussd'],
  ...
])
```

src/store/flow/block.ts
```diff
@@ -73,6 +73,17 @@ export const mutations: MutationTree<IFlowsState> = {
 }
 
 export const actions: ActionTree<IFlowsState, IRootState> = {
+  block_setLabel({commit}, {blockId, value}) {
+    const snakeCase = (value: string) => value?.trim()
+      .replace(/[^\w\s]+/g, '')
+      .split(' ')
+      .map((word) => word.toLowerCase())
+      .join('_')
```
bzabos, line 81: Curiosity Q: Does [lodash's snakeCase](https://lodash.com/docs/4.17.15#snakeCase) behave similar enough that we could use it here instead of hand-rolling?

src/components/common/SvgIcon.vue
```diff
@@ -0,0 +1,42 @@
+<template>
```
bzabos, line 1: These SVGIcon + SVGSprite components look good! I'm wondering if we could leverage fontawesome's library management to offload some of the work/complexity? 

I was personally curious and did a quick google search leading me to [a ts example](https://codesandbox.io/s/fa-list-old-style-ts-k3zhc?file=/src/index.ts) via https://github.com/FortAwesome/vue-fontawesome/issues/274#issuecomment-798897059. Would something like that help us or does it make things more complicated? ⚙️⚙️⚙️

bzabos (approved):
Hi @poojahpatil90! Nice one, this looks good! I have left a couple minor comments. 

One curiosity thing: The pink/primary colour on the design seems to be more pinky/fuchsia-y I'm guessing I missed some discussions, but I'm wondering if those should match!

bzabos:
Hi @poojahpatil90, I'm just looking a bit more closely at the design a bit closer and I noticed two things:

- [ ] There's a small edit button next to the code as a label which is what someone should click to edit?
- [ ] Block semantic label: I'm confused as to whether this should be showing up under categorization below or up by the block name— I'm seeing it in both places, but under categorization makes the most sense to me

poojahpatil90:
@asalami Please review the design for this PR.

src/components/common/AudioLibrarySearchField.vue
```diff
@@ -19,6 +10,15 @@
         @focus="activate"
         @blur="deactivate"
         @input="resetPagination">
+
+      <span class="input-group-prepend">
```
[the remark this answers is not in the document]
poojahpatil90: Done!

src/components/interaction-designer/resource-editors/ResourceEditor.vue
poojahpatil90: Good catch!  I swapped them

src/components/interaction-designer/resource-editors/ResourceEditor.vue
poojahpatil90: I think this way is better. I changed it!

src/store/flow/block.ts
poojahpatil90: Yes, that would work!  I did not realize lodash had a function for it, thanks for sharing 👍

src/components/common/SvgIcon.vue
poojahpatil90: Hey @bzabos!  This is a great solution. Unfortunately, it works with only a single path within an svg. Many of our custom SVG icons have multiple paths. I will check with Godspower and Amen if it is possible to combine them into a single path. If so then we can use this approach, I really liked this approach better.

poojahpatil90:
> Hi @poojahpatil90, I'm just looking a bit more closely at the design a bit closer and I noticed two things:
> 
> * [ ]  There's a small edit button next to the code as a label which is what someone should click to edit?
> * [ ]  Block semantic label: I'm confused as to whether this should be showing up under categorization below or up by the block name— I'm seeing it in both places, but under categorization makes the most sense to me

@bzabos As discussed on Friday, I seemed to have used an older version of the UX. I made changes to include the edit button and block semantic label.

poojahpatil90:
> Hi @poojahpatil90! Nice one, this looks good! I have left a couple minor comments.
> 
> One curiosity thing: The pink/primary colour on the design seems to be more pinky/fuchsia-y I'm guessing I missed some discussions, but I'm wondering if those should match!

Screen Shot 2021-07-11 at 1 04 23 PM
@bzabos  as seen in the image above, I had used the color on the left hand side which seems to be a darker shade than the one on right hand side. I am not sure which is the right color though 🤷🏽‍♀️. Good thing is that we can change it in just one place now (custom_variables.scss). I have changed it to the lighter shade. Let me know your thoughts.

src/components/common/SvgIcon.vue
poojahpatil90: @asalami Any thoughts on the possibility to combine multiple paths in an svg to a single path ?  For example: In the screenshot we see two paths under the svg. I need just one path, is this possible? 
<svg> 
<path d=""> 
</svg>
Screen Shot 2021-07-11 at 1 19 09 PM

src/components/interaction-designer/block-types/MobilePrimitives_MessageBlock.vue
```diff
@@ -52,13 +55,17 @@ const builderVuexNamespace = namespace('builder')
     BlockSemanticLabelEditor,
     FirstBlockEditorButton,
     BlockId,
+    TitleEditor,
```
jcbashdown, line 58: @poojahpatil90 We've just discussed this in the FLOIP meeting and to avoid confusion we want to ensure these editor components reflect the name in the spec. We then want to change the labels of these fields using localisation only. Can we change back Title Editor, Block Code Editor and anything else which is similar?

CC @bzabos

src/components/common/SmallTextEditor.vue
```diff
@@ -0,0 +1,44 @@
+<template>
```
safydy-r, line 1: @poojahpatil90 , I feel like adding a new component would increase maintenance complexity in the future. Eg: if we'd like to update style, or changing the way we use validation component, etc.

As the difference dwell in template only, I suggest we keep using the old `TextEditor` component:
- solution1: add [rows attribute](https://www.w3schools.com/tags/att_textarea_rows.asp) to text-area to handle the height
- solution2: make a conditional render using a new added prop `rows` if we really want an `<input>` tag. Pseudo code:
```
<text-area v-if='rows > 1' :rows="rows" .... />
<input v-else ... />
```

What do you think ?

src/lib/custom-icons/index.ts
```diff
@@ -0,0 +1,86 @@
+import {
```
safydy-r, line 1: @poojahpatil90 , I like this approach, thanks 🙏 💯

src/lib/custom-icons/index.ts
```diff
@@ -0,0 +1,86 @@
...
...85 25.2208 25.362 25.0368 25.4614 24.8049C25.6599 24.3417 25.923 23.684 26.1859 22.8953C26.7067 21.3329 27.25 19.1891 27.25 17H24.75ZM3.75 14.6154V18.9615H6.25V14.6154H3.75ZM6 21.2115H9.89621V18.7115H6V21.2115ZM9.75318 21.1666L13.7127 23.9286L15.143 21.8782L11.1835 19.1161L9.75318 21.1666ZM17.25 22.0832V11.0761H14.75V22.0832H17.25ZM13.5974 9.31673L9.70486 12.4199L11.2633 14.3747L15.1558 11.2716L13.5974 9.31673ZM9.8607 12.3654H6V14.8654H9.8607V12.3654ZM9.70486 12.4199C9.74914 12.3846 9.80408 12.3654 9.8607 12.3654V14.8654C10.3703 14.8654 10.8648 14.6924 11.2633 14.3747L9.70486 12.4199ZM17.25 11.0761C17.25 9.19123 15.0713 8.14179 13.5974 9.31673L15.1558 11.2716C14.9921 11.4021 14.75 11.2855 14.75 11.0761H17.25ZM13.7127 23.9286C15.2043 24.969 17.25 23.9018 17.25 22.0832H14.75C14.75 21.8811 14.9773 21.7625 15.143 21.8782L13.7127 23.9286ZM9.89621 21.2115C9.84506 21.2115 9.79513 21.1958 9.75318 21.1666L11.1835 19.1161C10.8059 18.8528 10.3566 18.7115 9.89621 18.7115V21.2115ZM3.75 18.9615C3.75 20.2042 4.75736 21.2115 6 21.2115V18.7115C6.13807 18.7115 6.25 18.8235 6.25 18.9615H3.75ZM6.25 14.6154C6.25 14.7535 6.13807 14.8654 6 14.8654V12.3654C4.75736 12.3654 3.75 13.3727 3.75 14.6154H6.25Z',
```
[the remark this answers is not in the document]
safydy-r: what if we suppress ESLint for this kind of line ?
`// eslint-disable-next-line max-len`

src/main.ts
```diff
@@ -17,13 +17,14 @@ import 'scss/main.scss'
...
-import {faCircle, faClone, faTrashAlt} from '@fortawesome/free-regular-svg-icons'
+import {faCheckCircle, faCheck, faMobileAlt} from '@fortawesome/free-solid-svg-icons'
+import {faCircle, faClone, faTrashAlt, faEdit, faEnvelope, faCommentDots} from '@fortawesome/free-regular-svg-icons'
 import {FontAwesomeIcon} from '@fortawesome/vue-fontawesome'
+import CustomIcons from './lib/custom-icons/index'
 import router from './router'
 import App from './App.vue'
 
-library.add(faCircle, faCheckCircle, faClone, faTrashAlt)
+library.add(faCircle, faCheckCircle, faEdit, faCheck, faEnvelope, faMobileAlt, faCommentDots, faClone, faTrashAlt, ...CustomIcons)
```
safydy-r, line 27: <3

src/components/interaction-designer/block-types/MobilePrimitives_MessageBlock.vue
safydy-r: Agree with Jacob here, 
Title is `label`, Code is `name` right ?

src/components/common/SvgIcon.vue
poojahpatil90: Fixed

src/components/interaction-designer/block-types/MobilePrimitives_MessageBlock.vue
poojahpatil90: Fixed!

@safydy-r you are right. Title is `label` and Code is `name`

src/main.ts
poojahpatil90: @safydy-r  FYI, We can use the icons from font awesome if they are identical to the design(confirmed by Amen). If not, we can add it to the custom icons.

src/lib/custom-icons/index.ts
poojahpatil90: Thanks 🙏

src/components/common/SmallTextEditor.vue
poojahpatil90: This is a good idea!  I have made the change. Made rows to have a default value of 1.

yeboaasare (approved):

PR #110, opened 2021-07-08
Vmo 4246 small tweaks for integration

jcbashdown:
Lots of little tweaks to improve the embedded builder. 

- Remove uuid from the persist path to match what George and I agreed
- Update routes docs
- Update the export panel to prevent editing and make clear it's for export only
- Add the ability to remove the note from the import page as the default not is not relevant to the embedded builder and we haven't yet agreed a note for that context
- Export messages.json to allow for easier overriding in consuming apps
- Fix the contents of the export text area to make it so that if you have a valid single flow in vuex, the json there will be valid for import without any editing.


src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -237,15 +238,8 @@ export default class TreeBuilderToolbar extends mixins(Routes, Permissions, Lang
...
-    const {
-      flows,
-      resources,
-    } = this as { flows: IFlow[], resources: IResource[] }
     return JSON.stringify(
-      {
-        flows,
-        resources,
-      },
+      omit(this.activeFlowContainer, 'isCreated'),
```
jcbashdown, line 242: The purpose of this change is to make sure you get a valid flow container from the export panel so we can potentially demo an export/import loop

src/views/ImportFlow.vue
```diff
@@ -285,4 +285,9 @@ export default ImportFlow
   overflow: hidden;
   white-space: nowrap;
 }
+
+.card {
```
jcbashdown, line 289: Quick fix to ensure that this is always display flex as it is in the community builder so that when you remove the import-note from messages.json you still get the same width card.

src/views/ImportFlow.vue
```diff
@@ -11,6 +10,7 @@
               </h2>
               <p>{{ 'flow-builder.create-flow-from-json' | trans }}</p>
               <div
+                v-if="trans('flow-builder.import-note')"
```
poojahpatil90, line 13: @jcbashdown  Can you help me understand why we are checking for the import-note's presence?  Is this a dynamically populated string?

src/views/ImportFlow.vue
poojahpatil90: Can we use the bootstrap `d-flex` class instead?

poojahpatil90 (approved):
@jcbashdown  Overall, looks good. Just left some minor comments.

src/views/ImportFlow.vue
jcbashdown: I'm not sure this is a great way to do it but basically, I'm trying to allow for this message being empty. When embedded in voto5 the default message doesn't make sense. The intention is to allow overriding with an empty string (and embedding app should be able to override and localisation) but the element is still rendered if this happens. See here:

https://bitbucket.org/voto/voto5/pull-requests/1899/vmo-4246-integration-test-branch#Lresources/assets/js/floip/builder/routes/index.jsT26

PR #115, opened 2021-07-13
Vmo 4212 block toolbar on hover

bzabos:
@poojahpatil90 @safydy Do you think this is a fair solution? I simply background:white + fade-in toolbar right over top of the block. I've inherited borders to make it easier to match up if we adjust styling in the future.

### DIFF
--- a/src/components/interaction-designer/Block.vue
+++ b/src/components/interaction-designer/Block.vue
@@ -17,7 +17,7 @@
       @dragStarted="selectBlock"
       @dragEnded="handleDraggableEndedForBlock"
       @destroyed="handleDraggableDestroyedForBlock">
-      <div class="d-flex justify-content-between" v-if="showToolbar">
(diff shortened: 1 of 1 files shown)


src/components/interaction-designer/Block.vue
```diff
@@ -622,6 +623,23 @@ export default {
...
   background-color 200ms ease-in-out;
 
+  .block-toolbar {
+    transition: opacity 200ms ease-in-out;
+    background: white;
+    opacity: 0; // default state of hidden
+
+    margin-top: -39.75px;
+    margin-right: -8px;
+    margin-left: -8px;
```
bzabos, line 633: I used negative margins to take what was inside and pull the whole block contents outside of the .block container.

poojahpatil90 (approved):
@bzabos  This looks great!  Thanks for working on this 👍  Just found a minor issues with the border of the block when the block is not active. 

Screen Shot 2021-07-13 at 1 30 36 PM

PR #116, opened 2021-07-14
VMO-4284 - change to fix toolbar width

jcbashdown:


bzabos (approved):
@jcbashdown I understand what we're shooting for here, thanks for finding something that will work for now! I don't have any alternative recommendations without diving into the resulting/computed CSS. 

This seems totally fine to me for now! 🙆‍♀️  it would be nice to move this out into more adaptable styles in the future if possible— Should we create a separate ticket fo that work?

jcbashdown:
Ticket for doing this a better way here and assigned to me:

https://viamoinc.atlassian.net/browse/VMO-4297

PR #118, opened 2021-07-16
Vmo 3917 add language on import conflicts resolved

jcbashdown:
Adds the ability to add language on import when there is a language mismatch.

Switched off in voto5 here:

https://bitbucket.org/voto/voto5/pull-requests/1946/vmo-3917-remove-add-language-button

We have decided for now to not implement this feature in voto5. Later, we will try to replace the page which adds languages on voto5 with the LanguageAdder component here. See these tickets:

https://viamoinc.atlassian.net/browse/VMO-4486

https://viamoinc.atlassian.net/browse/VMO-4487

**A note on the requirements for this:** 

It's slightly difficult now to trace back all the reasons for the exact form I've built for this as the DSN ticket doesn't seem to have a figma link and I did most of the work way back in June. The designs in DSN-103 are mostly hand drawn and there is some explanatory text in them which isn't fully fleshed out and I've omitted from my implementation. There are also parts of the designs that we agreed to drop after looking at the spec (e.g. a right to left language check box). I think the reason I use the ISO 3166 code though, even though the field isn't in the design, is because of this description of the BCP 47 from Amen in the DSN ticket:

- Description text IETF BCP 47: “Optionally, you can add a language variant (including gender) to describe a specialization. An example of a language variant could be “female_east_africa,”. With “female” being the gender and “east_africa” being the language region, which is added to the ISO 639-3 code”.

The spec here says that ISO 3166 is one possible format for regions:

https://www.rfc-editor.org/rfc/bcp/bcp47.txt

And so I give this as an option which is then saved as part of the BCP 47 (region is not in the flow spec itself)

If there is one obvious omission from this PR it's the explanatory text that is in the designs next to the fields but not fully fleshed out. I think this is okay to leave for now as we actually plan to make changes to this form when we use it in voto5 and we ideally want to get this into QA before Amen is back.


src/components/interaction-designer/flow-editors/import/LanguageAdder.vue
```diff
@@ -0,0 +1,253 @@
...
+  components: {
+    BModal,
+    ValidationMessage,
+    VueMultiselect,
+    TextEditor,
+  },
+})
+class LanguageAdder extends mixins(Lang, Routes) {
+  addLanguageError = ''
+  newLanguage = {
```
safydy-r, line 127: I think using the `ILanguage` interface to initiate an object would be cleaner:
`newLanguage = {} as ILanguage`

src/components/interaction-designer/flow-editors/import/LanguageAdder.vue
```diff
@@ -0,0 +1,253 @@
...
+  selected_iso_3166_1: any = {}
+  iso_639_3Tags: any[] = iso6393
+  // There is no obvious way to narrow down ISO 3166-1 from iso 639 languages:
+  // https://www.rfc-editor.org/rfc/rfc5646.html#section-4.2
+  // As a result, we don't attempt to filter this list depending on the iso 639 selection
+  // We use Alpha-2 Codes as apparently they are "the most widely used"
+  // https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
+  resetLanguage(): void {
+    this.newLanguage = {
+      id: '',
```
safydy-r, line 144: Same as above, can we initiate from ILanguage here ? `this.newLanguage = {} as ILanguage`

src/components/interaction-designer/flow-editors/import/LanguageAdder.vue
```diff
@@ -0,0 +1,253 @@
...
+@Component({
+  components: {
+    BModal,
+    ValidationMessage,
+    VueMultiselect,
+    TextEditor,
+  },
+})
+class LanguageAdder extends mixins(Lang, Routes) {
+  addLanguageError = ''
```
safydy-r, line 126: This is unused, right ?

src/components/interaction-designer/flow-editors/import/LanguageAdder.vue
```diff
@@ -0,0 +1,253 @@
...
+    }
+    this.selected_iso_639_3 = {}
+    this.validation_removeNewLanguageValidation()
+  }
+  showAddLanguageModal(): void {
+    this.resetLanguage()
+    const languageModal: any = this.$refs['add-language-modal']
+    languageModal.show()
+  }
+  customLanguageLabel(option: any): string {
```
safydy-r, line 158: Hum, I never knew we can customize `vue-multiselect` like this, interesting ;)

src/components/interaction-designer/flow-editors/import/LanguageAdder.vue
```diff
@@ -0,0 +1,253 @@
...
+  filterVariant(e: KeyboardEvent): void {
+    if (e.key.match(/[^a-z_]/g)) {
+      e.preventDefault()
+    }
+    const lastCharacter = this.newLanguage.variant[this.newLanguage.variant.length - 1]
+    if (lastCharacter === '_' && e.key.match(/_/g)) {
+      e.preventDefault()
+    }
+  }
+  // For now, we aren't allowing the use of 'script' or other elements in BCP 47 Construction - though the spec allows this.
```
safydy-r, line 213: Can we use the multi-line comment format here?
```
/**
*
*/
```

src/store/validation/index.ts
```diff
@@ -119,6 +119,22 @@ export const actions: ActionTree<IValidationState, IRootState> = {
     debugValidationStatus(state.validationStatuses[key], 'flow container validation status')
     return state.validationStatuses[key]
   },
+
+  async validate_new_language({state, rootGetters}, {language}: { language: ILanguage }): Promise<IValidationStatus> {
+    const validate = getOrCreateLanguageValidator(rootGetters['flow/activeFlowContainer'].specification_version)
+    const index = 'language/new_language'
```
safydy-r, line 125: I like this approach for new language validator, thanks 🙏 ☺️

src/store/trees/trees.js
```diff
@@ -387,6 +393,27 @@ export default {
...
 
   actions: {
+    async validateAndAddOrgLanguage({dispatch, commit}, {persistRoute, language}) {
+      const validationErrors = await dispatch('validation/validate_new_language', {language}, {root: true})
+      if (!validationErrors.isValid) {
+        return false
+      }
+
+      if (!persistRoute) {
+        console.info('Language persistence route not configured correctly in builder.config.json. Falling back to vuex store')
```
safydy-r, line 403: As this event means we're having wrong configuration, I'd suggest we use `console.warn` instead. The same for the try catch below.

src/components/interaction-designer/flow-editors/import/LanguageAdder.vue
```diff
@@ -0,0 +1,253 @@
...
+            :label="'flow-builder.label' | trans"
+            :valid-state="isValid" />
+        </validation-message>
+      </div>
+      <div class="form-group">
+        <!-- From https://www.npmjs.com/package/iso-639-3 and use search input from resource editor -->
+        <validation-message
+          #input-control="{ isValid }"
+          message-key="language/new_language/iso_639_3">
+          <label class="form-check-label mt-2 mb-2 mr-2">{{'flow-builder.iso-639-3-label' | trans}}</label>
```
safydy-r, line 30: And let's add `text-primary` class for these labels to have uniform styling with `<text-editor ` label:
- ISO 639 3 Code, ISO 3166 1 Locale (optional), 
- BCP 47 
- Generated from ISO 639 3, Locale, and Variant

safydy-r (changes requested):
@jcbashdown  , I went through the PR. Overall, I think it's good.

I left some minor concerns/suggestions.

scripts/prepare-gh-pages-deployment.sh
```diff
@@ -15,8 +15,8 @@ abort()
 *** - Regenerate patches:
 ***     $ git diff src/router/index.ts > patches/gh-pages-prepare-router.patch
 ***     $ git diff vue.config.js > patches/gh-pages-prepare-vue-config.patch
-***     $ git diff builder.config.js > patches/remove-backend-routes-from-config.patch
```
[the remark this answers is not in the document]
jcbashdown: I guess people were just manually fixing the copied command!

src/store/trees/trees.js
jcbashdown: I think when using the standalone builder it should be fine to not configure this route but it's still useful info to know you can if you want. I think I'm going to leave this as is.

safydy-r: ok, thanks for that extra clarification

PR #122, opened 2021-07-20
VMO-4140 Add prop for textarea # of rows on resource-variant-text-editor

bzabos:
Hi @poojahpatil90, I've merged the simplified choices but wanted to get this fix in there as well which allows us to provide # of rows to the resource variant textarea.


poojahpatil90 (approved):
Looks good to me! 
I had setup rows to have 1 row by default. I might have to make that change so that we have consistent code everywhere. I have noted this bit.

PR #124, opened 2021-07-20
Vmo 4306 make sidebar headers having same style

safydy-r:
**Note**:
I've take the opportunity to udpate sidebar for block types as well to have label having the same style. See diff for more details.

Screenshot for FlowEditor:
Screen Shot 2021-07-20 at 1 27 19 PM


src/components/interaction-designer/block-editors/LabelEditor.vue
```diff
@@ -16,7 +16,7 @@
           @click="emitGearClickedEvent">
           <font-awesome-icon
             :icon="['fac', 'settings']"
-            class="fa-btn text-primary" />
+            class="fa-btn" />
```
poojahpatil90, line 19: Thanks for this 👍

poojahpatil90 (approved):
Thanks for fixing these label inconsistencies 🥇  I pushed a small fix for Message and Numeric block titles to include the title color you are using.

PR #129, opened 2021-07-21
VMO-4130 - remove call to record button and update bundle

jcbashdown:


safydy-r (approved):
Thanks @jcbashdown for working on this

PR #132, opened 2021-07-21
Vmo 4317 fix resource config for duplicated block

safydy-r:
Duplicated block has now a cloned resource from the originating block:
Screen Shot 2021-07-21 at 1 14 01 PM

### DIFF
--- a/tests/unit/__snapshots__/storybook.spec.ts.snap
+++ b/tests/unit/__snapshots__/storybook.spec.ts.snap
@@ -2206,11 +2206,7 @@ exports[`Storyshots MobilePrimitives/Numeric Response Block Existing Data For Te
                                           </div>
                                         </div>
                                          
-                                        <div
-                                          class="phone-recorder"
-                                        >
-                                          <!---->
-                                        </div>
+                                        <!---->
                                          
                                         <div
                                           class="upload-monitor"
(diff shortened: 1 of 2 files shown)


tests/unit/__snapshots__/storybook.spec.ts.snap
```diff
@@ -2206,11 +2206,7 @@ exports[`Storyshots MobilePrimitives/Numeric Response Block Existing Data For Te
                                           </div>
                                         </div>
                                          
-                                        <div
```
safydy-r, line 2209: this should have been done in other PR (which was already merged)

poojahpatil90: @safydy-r  I think we don't want to remove this piece. But remove these features in `builder.config` and `appconfig` so that these buttons don't show up in the UI. This needs to be hidden in the community version, but seen when embedded in voto5. 
@bzabos  Can you confirm if my understanding is right?

safydy-r: @poojahpatil90 , this is just a snapshot. The test found we shouldn't add call-to-record based on permission from `builder.config`, so that's why we won't have it in snapshot ;D

poojahpatil90 (approved):
@safydy-r I am sorry, I did not notice that it was a snapshot. Approved!

PR #134, opened 2021-07-21
Vmo 4323 fix overlapping blocks

safydy-r:
The pbm was we put `block.ui_metadata...` rather than `ui_metadata...` in `lodash.get` path. I took the opportunity to remove the usage of `get()` in some part of the code.

Screenshots
Screen Shot 2021-07-21 at 3 02 17 PM


poojahpatil90 (approved):
Looks good to me

PR #135, opened 2021-07-21
VMO-4246 - don't create resource on exit create

jcbashdown:
In RC2 we no longer need to create resources when we create exits.


safydy-r (approved):
No comment, thanks @jcbashdown

PR #141, opened 2021-07-24
VMO-4162 add BlockOutputBranchingConfig

safydy-r:
Note:
- setting exit.value when having Single exit mode will be done in vmo-4322
- in master, we already have setContactProp & Categorization

Screenshot
Screen Shot 2021-07-23 at 6 04 16 PM


jcbashdown (approved):
Looks good

PR #143, opened 2021-07-24
Vmo 4149 unified block photo response

safydy-r:
**Note**:
- setting exit.value when having Single exit mode will be done in vmo-4322
- in master, we already have setContactProp & Categorization

About `prompt` implementation, I had this [thread](https://votomobile.slack.com/archives/CGGFKUP4Y/p1627083068197500) on squad2 channel to confirm it should be localized

Screenshots
Screen Shot 2021-07-23 at 7 46 25 PM

Screen Shot 2021-07-23 at 7 47 12 PM

### DIFF
--- a/src/components/interaction-designer/block-editors/PhotoPromptEditor.vue
+++ b/src/components/interaction-designer/block-editors/PhotoPromptEditor.vue
@@ -1,0 +1,84 @@
+<template>
+  <div class="photo-prompt-editor">
+    <label class="text-primary">{{ 'flow-builder.photo-response-prompt' | trans }}</label>
+    <div class="custom-control custom-checkbox">
+      <input
+        id="shouldIncludePrompt"
+        :value="shouldSetContactProperty"
+        :checked="shouldSetContactProperty"
+        type="checkbox"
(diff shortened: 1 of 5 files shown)


src/components/interaction-designer/block-editors/PhotoPromptEditor.vue
```diff
@@ -0,0 +1,84 @@
...
+@Component({
+  components: {
+    ResourceEditor,
+  },
+})
+class LabelEditor extends mixins(Lang) {
+  @Prop() readonly block!: IBlock
+  @Prop() readonly flow!: IFlow
+
+  shouldSetContactProperty = false
```
jcbashdown, line 46: I don't really understand why this is called `shouldSetContactProperty`. Can you explain?

safydy-r: Very nice catch @jcbashdown , I made some logic copy-paste from other component. It should have been `shouldIncludePrompt`. 
I made a fix, thanks

jcbashdown (approved):

PR #144, opened 2021-07-24
Vmo 4322 add test values based on exit mode

safydy-r:
**Part1**: implementation for Message, Numeric, SelectOne blocks

Main changes:
- moved `block_convertExitFormationToUnified()` from `BlockOutputBranchingConfig.vue` to each block-type store, because we will customize `test` per block-type according to vmo-4322 AC
- this PR concern only MessageBlock, NumericBlock, SelectOne
- I made small update on all `block-types/{...}.ts` store to fix storybook issue related to undefined `vendor_metadata`

Screenshot:
Screen Shot 2021-07-23 at 9 02 58 PM


bzabos (approved):
Hi @safydy-r ! This looks perfect. The only things I'd consider is moving the actual logic and manipulation of strings to the block's corresponding vuex stores and dispatch `block_convertExitFormationToUnified()` from there, instead of doing it in the block's vue file. Possible more reusable that way 👌

src/components/interaction-designer/block-editors/BlockOutputBranchingConfig.vue
```diff
@@ -160,7 +160,7 @@
 
       switch (value) {
         case OutputBranchingType.UNIFIED:
-          this.block_convertExitFormationToUnified({blockId})
+          this.$emit('branchingTypeChangedToUnified')
```
bzabos, line 163: I'm wondering if we should do both of these so that the wrapping component is responding to the change rather than expecting the wrapper component to commit the change. It feels a bit different than other editors in that it that the wrapper NEEDs to implement something. 

Is there a default scenario where we would want to be able to drop `<BlockOutputBranchingConfig />` in and not have to implement any handler?

bzabos: Then the wrapper component would receive the event and update the test property itself using its own block store? I'm not 100% confident, but what are your thoughts?

safydy-r: According what we have discussed, we will leave it as I have implemented.

PR #148, opened 2021-07-26
VMO-4146 - unified print block

jcbashdown:
Note - This diverges from the designs and removed the resource editor. [Mark has asked](https://votomobile.slack.com/archives/CGGFKUP4Y/p1627246325213900?thread_ts=1627083068.197500&cid=CGGFKUP4Y) us to allow a string only for now (which can be an expression)


safydy-r (approved):
No comment as well, ... approving & merging

PR #149, opened 2021-07-26
Vmo 4349 save gives feedback

jcbashdown:
- Remove save button on github pages - it does nothing. Running it standalone locally we have a fake backend so we can still have the button.
- Switch text on save button when saving, restyle button to give constant width
- Button is already disabled when saving
- Leaving change detection for now - we need a solution that works throughout the relevant bits of the vuex state. Save will always be available even without any further changes. Can you think of any quick way to achieve this though? If so, we can create a further ticket to try and tackle it this week.

Demo of this working in voto5 with real backend.

https://user-images.githubusercontent.com/677893/127033414-0db99aa1-f315-4627-b2a0-f5a27cc7f903.mp4


scripts/prepare-gh-pages-deployment.sh
```diff
@@ -16,6 +16,7 @@ abort()
 ***     $ git diff src/router/index.ts > patches/gh-pages-prepare-router.patch
 ***     $ git diff vue.config.js > patches/gh-pages-prepare-vue-config.patch
 ***     $ git diff builder.config.js > patches/remove-backend-routes-from-config.patch
+***     $ git diff > patches/disable-save-button.patch
```
safydy-r, line 19: Thanks, I like this suggestion

safydy-r (approved):
No concerns here, ... approving

PR #154, opened 2021-07-27
VMO-4346 - fix exit becoming invalid value on create without target

jcbashdown:
Setting to null when you don't finish connecting a block is invalid - unset completely.

Update some example flows as currently the config for these test flows is missing some important exit defaults we have added recently. Without these if you play with exits without viewing the block sidebar first then things can behave strangely.

I wasn't able to reproduce the first problem with this ticket. I've asked Jon for further instructions. Let's try and get this one in and I'll do that in a separate PR.


safydy-r (approved):
No comment here. This is a neat fix I think

PR #155, opened 2021-07-27
VMO-4345 Fix when toolbar width inconsistent w/ multiple exits.

bzabos:


poojahpatil90 (approved):
@bzabos  Thanks for working on this it was really tricky to get it done!

PR #156, opened 2021-07-27
VMO-4255 - isolate styles

jcbashdown:


bzabos (approved):
Nice one @jcbashdown!  This is so much simpler than what I was proposing 🤩!

PR #157, opened 2021-07-27
Vmo 4322 add test values based on exit mode - part2

safydy-r:
Part2: Implementation for
- Print
- Read
- Output
- RunFlow
- SetContactProp
- SetGroupMembership
- OpenResponse
- LocationResponse
- PhotoResponse

(In case of need, the PR for part1: https://github.com/FLOIP/flow-builder/pull/144)

In part3 (next PR):
- Log (waiting for Jacob's VMO-4148)
- Case (waiting for Jacob's VMO-4161)
- SelectMany (waiting for Brett's VMO-4141) & confirmation from Mark https://votomobile.slack.com/archives/CMQDVRDN3/p1627395757103400?thread_ts=1627339322.102400&cid=CMQDVRDN3


safydy-r:
I'll fix the build after approval, it 's a simple snapshot update

bzabos (approved):
Hi @safydy, thanks! I think this looks good! 

Side-note: I still feel a bit weird about not having a default handler for the component itself (especially when many end up having `{test: 'block.value = true'}`), but I think the pattern you've laid out is nice in that each store covers their own business. 😻🙆‍♀️

PR #160, opened 2021-07-27
Vmo 4383 update block config title

safydy-r:
I took the opportunity to allow us having null value on ExpressionInput, so we can see validation if we provide `null` value for example.

### DIFF
--- a/src/components/interaction-designer/block-types/SmartDevices_PhotoResponseBlock.vue
+++ b/src/components/interaction-designer/block-types/SmartDevices_PhotoResponseBlock.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="smart-devices-photo-response-block">
     <h3 class="no-room-above">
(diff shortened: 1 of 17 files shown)


src/components/interaction-designer/block-types/SmartDevices_PhotoResponseBlock.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="smart-devices-photo-response-block">
     <h3 class="no-room-above">
-      {{ 'flow-builder.edit-block-type' | trans({block_type: trans(`flow-builder.${block.type}`)}) }}
+      {{ `flow-builder.${block.type}` | trans }}
```
poojahpatil90, line 4: Thanks for changing this for all block types 👍

poojahpatil90 (approved):
Everything looks good to me. Approved

PR #161, opened 2021-07-28
VMO-4330 add temporary solution to fix vertical scroll

safydy-r:
This is a temporary solution, but we should review it when we're done with toolbar update.
I will add a TODO in https://viamoinc.atlassian.net/browse/VMO-4192 to remind us.

Overview after fix:
vmo4330

### DIFF
--- a/src/components/interaction-designer/BuilderCanvas.vue
+++ b/src/components/interaction-designer/BuilderCanvas.vue
@@ -32,7 +32,7 @@ const MARGIN_HEIGHT = 100
 const MARGIN_WIDTH = 100
 
 //ms
-const DEBOUNCE_SCROLL_TIMER = 100
+const DEBOUNCE_SCROLL_TIMER = 300
 
 @Component({
   components: {
@@ -116,7 +116,7 @@ export default class BuilderCanvas extends Vue {
         `block/${this.blockAtTheLowestPosition?.uuid}`,
       )
       // temporary dummy height for UI scroll purpose
-      return 150
+      return 84
     }
     return (<HTMLElement>(<Vue>blockElementRef[0].$refs.draggable).$el).offsetHeight
   }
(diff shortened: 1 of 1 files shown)


src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -165,7 +165,9 @@ export default class BuilderCanvas extends Vue {
     }
 
     const yPosition: number = this.blockAtTheLowestPosition.ui_metadata.canvas_coordinates.y
-    const scrollHeight = yPosition + this.blockHeight + MARGIN_HEIGHT
+    // TODO: clean this original solution in case the actual solution still works after we update the toolbar. I suspect there is a relation between the scrollHeight formula and the fact that we have toolbar elements with fixed position
```
safydy-r, line 168: I added a todo in https://viamoinc.atlassian.net/browse/VMO-4192 to remind us about this

bzabos:
Hi @safydy I don't think I quite understand what this changeset is addressing or even what the changes effect 😹  Can we connect briefly on this tomorrow morning?

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -116,7 +116,7 @@ export default class BuilderCanvas extends Vue {
         `block/${this.blockAtTheLowestPosition?.uuid}`,
       )
       // temporary dummy height for UI scroll purpose
-      return 150
+      return 84
```
safydy-r, line 119: I will create a constant here

bzabos (approved):
Hi @safydy, after our discussion this is now clear to me. Thanks! Looks good.

src/components/interaction-designer/BuilderCanvas.vue
bzabos: I think we want to use the convention of either erasing the todo or putting the JIRA number right inside the code's todo comment so other people can refer to it and add details if they need to! Would you be willing to add the JIRA number into the code comment? Thanks!

PR #162, opened 2021-07-28
VMO-4328 do not remove exits draggable for view mode

safydy-r:
Solution: We should not hide exit draggables on view mode, because connections are bound with them.

[After fix]
vmo4328


jcbashdown (approved):
Looks good.

PR #163, opened 2021-07-28
VMO-4365 Remove error exits + fix SelectMany's exits for consistent behaviour and spec compliance

bzabos:
Hi @safydy, I think this is right— The LogBlock doesn't have any branching configuration so it presents with Default on initi, and CaseBlock should be handled by @jcbashdown when he implements that task?

---

Screen Shot 2021-07-27 at 9 01 50 PM


src/store/flow/block-types/Core_SetContactPropertyStore.ts
```diff
@@ -23,9 +23,8 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
       await dispatch('flow/block_createBlockDefaultExitWith', {
         props: ({
           uuid: await (new IdGeneratorUuidV4()).generate(),
-          name: 'Default',
-        }),
-      }, {root: true}) as IBlockExit,
+        }) as IBlockExit,
```
safydy-r, line 26: haha, good spot here

src/store/flow/block-types/MobilePrimitives_SelectManyResponseBlockStore.ts
```diff
@@ -27,23 +27,20 @@ export const actions: ActionTree<ICustomFlowState, IRootState> = {
...
     const blankPromptResource = await dispatch('flow/flow_addBlankResourceForEnabledModesAndLangs', null, {root: true})
-    const defaultExitProps: Partial<IBlockExit> = {
-      uuid: await (new IdGeneratorUuidV4()).generate(),
-      name: 'Default',
-      default: true,
-      // test: '',
-    }
-
-    const defaultExit = await dispatch('flow/block_createBlockDefaultExitWith', { props: defaultExitProps }, { root: true })
+    const exits: IBlockExit[] = [
```
safydy-r, line 30: 👍

safydy-r (approved):
Approved.

About the 02 remaining blocks (Log & Case):
I think we should make sure Log block reflects these change too, cc: @jcbashdown (a PR was created for it, haven't looked at it yet, maybe it's already there though).

About spec:
We should talk to Mark/Jon to update the spec about `Default` exit by correcting the number of exit information in the doc.

safydy-r:
@bzabos , I saw Jacob has merged CaseBlock, if I'm not wrong we should remove `name` & `test` from these lines after you update from `master` https://github.com/FLOIP/flow-builder/pull/165/files#diff-ff45e46d56304ec0128fe3886bc1b84c897a154c43804678b92bc5889a3fadabR57, right ?

safydy-r:
And Log block has been merged too, so could you pls update its store to reflect your change ? Removing `name` here I think
https://github.com/FLOIP/flow-builder/pull/166/files#diff-ccd25eb2841c2996da14840919d28d0b54309dcb7b815fde655de465976df088R29

PR #164, opened 2021-07-28
VMO-4367 + VMO-4362 Fix when duplicating multiple blocks doesn't merge metadata

bzabos:
This fixes Default showing up on multi-duplicate.

---

Screen Shot 2021-07-27 at 10 09 19 PM

### DIFF
--- a/src/store/flow/flow.ts
+++ b/src/store/flow/flow.ts
@@ -16,7 +16,7 @@ import {IdGeneratorUuidV4} from '@floip/flow-runner/dist/domain/IdGeneratorUuidV
 import moment from 'moment'
 import {ActionTree, GetterTree, MutationTree} from 'vuex'
 import {IRootState} from '@/store'
-import {cloneDeep, defaults, every, forEach, get, has, includes, omit} from 'lodash'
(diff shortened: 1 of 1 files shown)


src/store/flow/flow.ts
```diff
@@ -361,14 +361,15 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
 
     // Set UI positions
     // TODO: remove this once IBlock has vendor_metadata key
```
safydy-r, line 363: @bzabos , I think we missed to remove this TODO, can we take the opportunity to remove this ?

src/store/flow/flow.ts
```diff
@@ -361,14 +361,15 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
-    duplicatedBlock.vendor_metadata = {
+    merge(duplicatedBlock.vendor_metadata, {
```
safydy-r, line 364: Neat solution, thanks 💯

safydy-r (approved):
Thanks @bzabos , very clear solution.
I left one minor request

src/store/flow/flow.ts
bzabos: 😻 I liked how clean this changeset was!

PR #165, opened 2021-07-28
Vmo 4161 unified case block

jcbashdown:
Includes a fix for a CSS compilation warning


bzabos (approved):
Thanks @jcbashdown!  I think this looks good 👌 !

PR #166, opened 2021-07-28
VMO-4148 - unified log block

jcbashdown:
In line with the design, we are making this only work with a single expression/string for now rather than a resource.


poojahpatil90 (approved):
@jcbashdown Everything looks good to me. Approved!

src/store/flow/block-types/Core_LogBlockStore.ts
```diff
@@ -12,13 +12,11 @@ export const getters: GetterTree<IFlowsState, IRootState> = {}
 
 export const mutations: MutationTree<IFlowsState> = {}
 export const actions: ActionTree<IFlowsState, IRootState> = {
-
+  async editMessage({commit}, {blockId, message}: { blockId: string, message: string }): Promise<string> {
+    commit('flow/block_updateConfig', {blockId, newConfig: {message}}, {root: true})
```
bzabos, line 16: This seems dangerous to me that we overwrite the entire config instead of modifying the single key the method indicates! Can we change this to only modify the message key instead? I'm just imagining that as a pattern, this would become a bug if we tried to add any other properties to the LogBlock's config interface. What do you think?

bzabos (approved):
Hi @jcbashdown!  I left a note about overwriting config vs. manipulating a single key, otherwise happy to merge 👌

src/components/interaction-designer/block-types/Core_LogBlock.vue
```diff
@@ -76,20 +73,23 @@ import FirstBlockEditorButton from '../flow-editors/FirstBlockEditorButton.vue'
...
-const flowVuexNamespace = namespace('flow')
+const blockVuexNamespace = namespace(`flow/${BLOCK_TYPE}`)
 const builderVuexNamespace = namespace('builder')
 
 @Component({
   components: {
     GenericContactPropertyEditor,
+    ExpressionInput,
     ResourceEditor,
     ResourceVariantTextEditor,
```
bzabos, line 84: Hi @jcbashdown just realized, we could also probably remove those two resource editor components 🤩

src/store/flow/block-types/Core_LogBlockStore.ts
jcbashdown: Good point. Fixed!

src/components/interaction-designer/block-types/Core_LogBlock.vue
jcbashdown: Removed these and a couple of other extraneous components too.

PR #167, opened 2021-07-28
VMO-4210 Keep the block editor's header sticky to the top

poojahpatil90:
@safydy-r I used gif here instead of mov files as you suggested 👍  Thanks for sharing this, it is so much lighter!

### DIFF
--- a/src/components/interaction-designer/block-editors/BlockEditor.vue
+++ b/src/components/interaction-designer/block-editors/BlockEditor.vue
@@ -55,9 +55,19 @@ export default {
   width: 365px;
   overflow-y: scroll;
   padding: 1em;
+  padding-top: 0;
   border: 1px solid lightgray;
   border-radius: 0;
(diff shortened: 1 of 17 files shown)


src/components/interaction-designer/block-editors/BlockEditor.vue
```diff
@@ -55,9 +55,19 @@ export default {
...
   padding: 1em;
+  padding-top: 0;
   border: 1px solid lightgray;
   border-radius: 0;
   box-shadow: 0 3px 6px #cacaca;
   transition: right 200ms ease-in-out, 200ms background-color ease-in-out, 200ms border-color ease-in-out, 200ms border-radius ease-in-out;
 }
+
+.block-editor-header {
+  position: sticky;
```
safydy-r, line 66: 💯 , thanks for investigating on this solution @poojahpatil90 .

safydy-r (approved):
Awesome, I like it, thanks

PR #168, opened 2021-07-29
VMO-4314 New block exits UX

poojahpatil90:
Note: This PR also fixes VMO-4363, VMO-4344. I had to combine these PRs as they were related.
There is a css issue with the block exit text being pushed a bit to the left and not aligning to the center once a connection is created. I will try to fix this issue soon and update the PR


src/components/interaction-designer/Block.vue
```diff
@@ -131,87 +132,80 @@
...
-                    class="glyphicon glyphicon-move" />
-                  <font-awesome-icon
-                    v-else
-                    class="text-danger"
-                    :icon="['far', 'dot-circle']" />
-                </plain-draggable>
-
-                <template v-if="isConnectionCreateActive && isExitActivatedForCreate(exit) && livePosition">
+            <div
+              class="block-exit-tag badge badge-warning d-flex justify-content-center"
```
bzabos, line 140: Can we rename `block-exit-tag` to `block-exit-name` ? 🏷️

src/components/interaction-designer/Block.vue
```diff
@@ -131,87 +132,80 @@
...
+              :style="{background: exitBackgroundColor(exit)}"
```
bzabos, line 141: Personally, I think these templates are clearest if we can take the approach of applying the semantic "state" of things using CSS classes, then implement the details of styling in the CSS. This gives us very clean [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns), keeping CSS + HTML out of JS and JS out of template!

An example (pls bear with the pseudo-code 😛 ):
```
<template>
...
    <div
      v-for="(exit, key) in block.exits"
      :key="exit.uuid"
      class="block-exit"
      :class="{
        ['is-hovered']: isExitHovered(exit),
        ['is-not-hovered']: !isExitHovered(exit),
        ['is-connected']: doesExitHaveConnection(exit),
        ['is-disconnected']: !doesExitHaveConnection(exit),
      <div 
        class="block-exit-tag ..."
        ... >
          {{ exit.name || '(untitled)' }}
         ...
</template>
```

```
<script>
...
methods: {
  isExitHovered(exit) {
    return exit.destination_block != null && (this.exitHovers[exit.uuid] || this.lineHovers[exit.uuid])
  },

  doesExitHaveConnection(exit) {
    return exit.destination_block != null
  }
  ...
}
</script>
```

```
<style>
.block {
  .block-exit {
    // default styling
    background: #xxx;

    // state cascaded styling
    &.is-connected {
      background: #xxx;
    }

    &.is-disconnected {
      background: #xxx;      
    }

    &.is-hovered {
      background: #xxx;      
    }
  }
}
<style>
```

---

Sorry for the verbosity— what do you think of this approach?

src/components/interaction-designer/Block.vue
```diff
@@ -131,87 +132,80 @@
...
+
+                <template v-if="exit.destination_block != null">
                   <div
                     :id="`exit/${exit.uuid}/handle`"
-                    class="block-handle-move btn btn-xs">
-                    <i class="glyphicon glyphicon-move" />
+                    :key="`exit/${exit.uuid}/handle`"
+                    class="btn btn-xs btn-flat">
+                    <font-awesome-icon
+                      v-if="exitHovers[exit.uuid]"
```
bzabos, line 190: No longer affected by isEditable? 🙀

src/components/interaction-designer/Block.vue
```diff
@@ -237,6 +231,10 @@ Vue.component('BTooltip', BTooltip)
 
 const LABEL_CONTAINER_MAX_WIDTH = 650
 
+const EXIT_CONNECTED_BG = '#418BCA'
+const EXIT_DISCONNECTED_BG = '#858585'
+const EXIT_HOVER_BG = '#FFECEC'
```
bzabos, line 236: We can remove these if we go with approach suggested above!

src/components/interaction-designer/Block.vue
```diff
@@ -743,23 +754,16 @@ export default {
       transition: border-radius 200ms ease-in-out;
 
       .block-exit-tag {
```
bzabos, line 756: Would need to be renamed if we rename to `block-exit-name` as mentioned above 👌

src/components/interaction-designer/Connection.vue
```diff
@@ -181,16 +182,17 @@ export default {
...
       this.line.setOptions(this.prominentOptions)
       this.activateConnection({connectionContext: this.connectionContext})
+      this.$emit('isLineHovered', true)
     },
     mouseOutHandler() {
       if (!this.isPermanentlyActive) {
         this.line.setOptions(this.options)
         this.deactivateConnection({connectionContext: this.connectionContext})
       }
+      this.$emit('isLineHovered', false)
```
bzabos, line 192: Do you think we should rename these? I found it a bit confusing at first glance in the consuming Block vue file because they're named as boolean flags rather than verbs/actions. 

An idea is to have two separate events matching convention of in/out: `lineMouseIn` + `lineMouseOut`— This also mitigates the need to provide the boolean. Thoughts?

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -602,19 +602,19 @@ export default class TreeBuilderToolbar extends mixins(Routes, Permissions, Lang
 
 <style lang="scss">
 .tree-builder-toolbar {
-  position: sticky;
```
bzabos, line 605: But I thought `sticky` was our newest bff!! 👯

src/views/InteractionDesigner.vue
```diff
@@ -2,9 +2,11 @@
   <div
     v-if="activeFlow"
     class="interaction-designer-contents">
-    <tree-builder-toolbar @height-updated="handleToolBarHeightUpdate" />
```
bzabos, line 5: Do you know if we need to then update CustomInteractionDesigner at all? 🤷‍♀️

bzabos (approved):
Hi @poojahpatil90, this looks great! Just a couple recommendations for an event rename and an approach to mitigate CSS in our JS+template. Thanks!

src/components/interaction-designer/Block.vue
poojahpatil90: Done!

src/components/interaction-designer/Block.vue
poojahpatil90: I like this approach!  This solved the issue of text contrast with the button too 😆  Thanks

src/components/interaction-designer/Block.vue
poojahpatil90: Done!

src/components/interaction-designer/Block.vue
poojahpatil90: Done!

src/components/interaction-designer/Connection.vue
poojahpatil90: I think renaming makes sense and made these changes.

src/components/interaction-designer/Block.vue
poojahpatil90: isEditable was removed on the master branch, I assumed it was removed to fix some bugs. But I notice that the icons come up on hover now. I figured a better way to deal with this by disabling all mouse events on the whole footer (might work on the root div of Block.vue too). Let me know your thoughts on this approach?

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
poojahpatil90: This was removed as it was redundant code that came in from Saf's changes. I have used this style on `interaction-designer` rather than using it in the Toolbar to wrap it around the whole Toolbar rather than pieces of it.

src/views/InteractionDesigner.vue
poojahpatil90: I am not sure about this 
@jcbashdown  Any thoughts of whether I need to make this change on `CustomInteractionDesigner` as well?

jcbashdown: Yes, we'd have to make changes to the interaction designer. Should be fairly simple though? I will look into doing this now in the 4193 PR

PR #170, opened 2021-07-29
Vmo 4141 select many min max choice editors

bzabos:
Hi @safydy, there's a lot of duplication between our numeric editors and these editors, but I think this is the pattern we'd agreed on 🤷‍♀️ .

Hi @poojahpatil90, I'd added some transparency to your sticky block editor headers to indicate that a user has scrolled up a bit, also moved tooltips to bottom for bottom actions to avoid covering so much of the block UI— Let me know what you think. 

---

Screen Shot 2021-07-28 at 10 19 04 PM

---

Screen Shot 2021-07-29 at 12 09 03 AM
Screen Shot 2021-07-29 at 12 09 12 AM


src/store/flow/block-types/MobilePrimitives_SelectManyResponseBlockStore.ts
```diff
@@ -44,13 +69,44 @@ export const actions: ActionTree<ICustomFlowState, IRootState> = {
...
+        minimum_choices: undefined,
+        maximum_choices: undefined,
       },
       vendor_metadata: {},
       tags: [],
     })
   },
 }
 
+function formatTestValueForUnifiedBranchingType(block: ISelectManyResponseBlock): string {
```
safydy-r, line 81: Thanks for adding this as part of the PR

src/store/flow/block-types/MobilePrimitives_SelectManyResponseBlockStore.ts
```diff
@@ -44,13 +69,44 @@ export const actions: ActionTree<ICustomFlowState, IRootState> = {
       config: {
         prompt: blankPromptResource.uuid,
         choices: {},
+        minimum_choices: undefined,
```
safydy-r, line 72: Can you give me an update when these new prop should be available from flow-runner ?

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
```diff
@@ -134,7 +122,7 @@ function formatTestValueForUnifiedBranchingType(block: ISelectOneResponseBlock):
     console.warn('Choices are empty for SelectOneBlock, providing `true` by default')
     return 'true'
   }
-  return `OR(${map(blockChoicesKey, (choice) => `block.value = '${choice}'`).join(',')})`
+  return `OR(${map(blockChoicesKey, (choice) => `block.value = "${choice}"`).join(',')})`
```
safydy-r, line 125: I think we should keep singe quote here to avoid escaping the character.
Mark said in [vmo-4322 description](https://viamoinc.atlassian.net/browse/VMO-4322)
> (Escaping of double-quotes as needed within JSON. You might be able to use single-quotes instead of double-quotes for surrounding the choice names)

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
```diff
@@ -88,7 +76,7 @@ export const actions: ActionTree<ICustomFlowState, IRootState> = {
 
       Object.assign(exitsForChoices[i], {
         name: choiceKey,
-        test: `block.value = ${i + 1}`,
+        test: `block.value = "${choiceKey}"`,
```
safydy-r, line 79: Related to the `simple quote` comment below

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
```diff
@@ -59,21 +55,13 @@ export const actions: ActionTree<ICustomFlowState, IRootState> = {
...
     Vue.delete(block.config.choices, choiceKey)
   },
 
   async reflowExitsFromChoices({dispatch, rootGetters}, {blockId}: {blockId: IBlock['uuid']}) {
-    const block: ISelectOneResponseBlock = rootGetters['builder/blocksById'][blockId]
-    if (block == null) {
-      throw new ValidationException(`Unable to find block: ${blockId}`)
-    }
-
+    const block: ISelectOneResponseBlock = findBlockWith(blockId, rootGetters['flow/activeFlow']) as ISelectOneResponseBlock
```
safydy-r, line 64: @bzabos , I curious if you didn't have issue on SelectMany when this action `reflowExitsFromChoices` is called by `handleChoiceChanged` or `reflowExitsWhenSwitchingToBranchingTypeNotUnified`?

I'm wondering if we shouldn't use `ISelectOneResponseBlock | ISelectManyResponseBlock` instead of `ISelectOneResponseBlock` only ? (applicable for the whole `as ISelectOneResponseBlock` & `:ISelectOneResponseBlock` iteration)

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
```diff
@@ -1,9 +1,11 @@
 import {ActionTree, GetterTree, MutationTree} from 'vuex'
```
safydy-r, line 1: @bzabos , I've request for you:
Could you in this particular line https://github.com/FLOIP/flow-builder/blob/VMO-4141-select-many-min-max-choice-editors/src/components/interaction-designer/block-types/MobilePrimitives_SelectOneResponseBlock.vue#L119 something like
```
    if (metadata.io_viamo.branchingType === UNIFIED) {
      this.handleBranchingTypeChangedToUnified({block: this.block})
    }
```
to make sure test values will be updated each time we have choices changes on UNIFIED branch type.

safydy-r (changes requested):
Thanks @bzabos for adding test values update in this PR 🙏 

I have some questions & requests for you in comments.

package.json
```diff
@@ -11,6 +11,7 @@
     "serve": "vue-cli-service serve",
     "build-bundle": "vue-cli-service build --target lib --name flow-builder src/lib.ts",
     "test:unit": "vue-cli-service test:unit",
+    "test:regenerate-storyshots": "vue-cli-service test:unit -u",
```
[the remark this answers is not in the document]
bzabos: Hi hi, any args we pass to `yarn test:regenerate-storyshots` will get forwarded onto the underlying call. Does that help? Like, if I did the following, the `vue-cli-service would receive the no-cache option:
```
$ yarn test:regenerate-storyshots --no-cache
```

With that in mind, do you think it would be fair to have just one command and we manually supply the additional flag when we need it?

src/components/interaction-designer/block-editors/MaximumChoicesEditor.vue
```diff
@@ -0,0 +1,61 @@
...
+  components: {
+    NumericEditor,
+    ValidationMessage,
+  },
+})
+export class MaximumChoicesEditor extends mixins(Lang) {
+  @Prop() readonly block!: ISelectManyResponseBlock
+
+  mounted(): void {
+    // todo: dispatch default value onto block
```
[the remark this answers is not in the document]
bzabos: I'll pick this up in my VMO-4370 ticket !

src/components/interaction-designer/block-editors/MaximumChoicesEditor.vue
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  // eslint-disable-next-line line-comment-position
+  get value(): number | string { //ISelectManyResponseBlockConfig['minimum_choices'] {
+    const max = this.block.config.maximum_choices
+    return isNumber(max) ? max : ''
```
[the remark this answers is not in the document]
bzabos: 🙀💔 My heart says we should try our very best to only ever manage and rely on a single type and consider it a bug if it's being casted to something else. I've got the `.number` modifier on the input, so it should be casted as a number on the way into this v-model setter. Do you think that's sufficient?

src/components/interaction-designer/block-editors/MaximumChoicesEditor.vue
```diff
@@ -0,0 +1,61 @@
...
+})
+export class MaximumChoicesEditor extends mixins(Lang) {
+  @Prop() readonly block!: ISelectManyResponseBlock
+
+  mounted(): void {
+    // todo: dispatch default value onto block
+  }
+
+  // eslint-disable-next-line line-comment-position
+  get value(): number | string { //ISelectManyResponseBlockConfig['minimum_choices'] {
```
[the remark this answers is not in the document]
bzabos: Okay, I'm cool with that— Is this a convention that we're using? Should it match the prop identically or can it be camel-cased/abbreviated? `maxChoices`?

src/store/flow/block-types/MobilePrimitives_SelectManyResponseBlockStore.ts
bzabos: We've got a JIRA scheduled and can discuss at huddle tomorrow! See: https://votomobile.slack.com/archives/CGGFKUP4Y/p1627585834423100

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
bzabos: Ok ok, my thinking on the spot was that we're already sluggifying choice names. So if the choice is `Cat's Korner`, we end up with `cats_korner` in the expression. However, I'm now realizing that we're not the only flow builder that will ever exit! 😹 

To address what you've mentioned, I'm a bit confused: I thought double quotes would be safer? Eg. the following would _not_ break, but with single quotes it would. Am I hearing the Case* you're trying to make properly?
```
block.value = "Cat's Korner"
```

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
bzabos: I didn't have any issues! The typing here still works because both have `.choices` property. I think it feels a bit weird to have SelectMany typing in the SelectOne store. Would you be okay if I left it as is?

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
bzabos: Ah! You're brilliant. I saw this bug at the end of the night last night, but was too tired to fix it 😛

package.json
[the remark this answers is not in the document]
safydy-r: Aha, in that case, let's have `"vue-cli-service test:unit --no-cache"`, like what we have in travis config https://github.com/FLOIP/flow-builder/blob/master/.travis.yml#L8

`-u` shouldn't always be used for test

src/store/flow/block-types/MobilePrimitives_SelectManyResponseBlockStore.ts
safydy-r: Thanks

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
safydy-r: If you don't have issues, let's keep it as it is. Thanks

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
safydy-r: You make a point for this `Cat's Korner` example.
I'm just wondering if why we shouldn't escape the double quote here? (as Mark mentioned in vmo-4322 description)

To be honest, I don't really have an idea for now which one is the correct formatting. But, if you're good with double quote, I'm fine. And maybe we need to use double quote for `choices` here https://github.com/FLOIP/flow-builder/pull/170/files#diff-54364b669a225aca3b26fa98446e1ca87c32c769a4db1e764a06e447eff24f60R93 as well.

package.json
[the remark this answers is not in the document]
bzabos: Done!

src/components/interaction-designer/block-editors/MaximumChoicesEditor.vue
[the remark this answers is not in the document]
bzabos: Did `minChoices`/`maxChoices`, let me know if I'm missing the point!

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
bzabos: Done!

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
bzabos: Ok ok, I've updated the other `choices` to be double-quoted.

src/store/flow/block-types/MobilePrimitives_SelectManyResponseBlockStore.ts
```diff
@@ -25,6 +26,44 @@ export const mutations: MutationTree<ICustomFlowState> = {
...
+      blockId: block.uuid,
+      key: 'maximum_choices',
+      value: isNumber(value) ? value : undefined,
+    }, {root: true})
+
+    // todo: we should probably review our getters + setters in vue files and pull domain logic our stores ?! Schedule this?11
+    const {vendor_metadata: metadata} = block as unknown as IBlockWithBranchingType
+    const {UNIFIED} = OutputBranchingType
+    if (metadata.io_viamo.branchingType === UNIFIED) {
+      dispatch('handleBranchingTypeChangedToUnified', {block})
```
bzabos, line 56: Hi @safydy, I've opted to drop these in at the vuex layer. I believe that we should actually be constraining our domain logic at the vuex layer more often rather than complex-ifying vue component relationships by `$emit()`ting events— I think we're all guilty of this, including myself 😉. This way if anyone else tries to modify the property with `setMaxChoices()`, they'll also magically get their exits updated appropriately.

Do you happen to agree with this moving forward?

src/components/interaction-designer/block-editors/MaximumChoicesEditor.vue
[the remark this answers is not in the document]
safydy-r: Got this `rely on a single type`, I'll consider that in the future

src/components/interaction-designer/block-types/MobilePrimitives_SelectOneResponseBlock.vue
```diff
@@ -115,7 +115,11 @@ export class MobilePrimitives_SelectOneResponseBlock extends mixins(Lang) {
 
   handleChoiceChanged(): void {
     const {uuid: blockId, vendor_metadata: metadata} = this.block as unknown as IBlockWithBranchingType
-    const {EXIT_PER_CHOICE} = OutputBranchingType
+    const {EXIT_PER_CHOICE, UNIFIED} = OutputBranchingType
+
+    if (metadata.io_viamo.branchingType === UNIFIED) {
+      this.handleBranchingTypeChangedToUnified({block: this.block})
```
safydy-r, line 121: Thanks for this 🙏

src/store/flow/block-types/MobilePrimitives_SelectManyResponseBlockStore.ts
safydy-r: @bzabos , great idea for moving it in vuex. Will keep in mind

src/store/flow/block-types/MobilePrimitives_SelectOneResponseBlockStore.ts
safydy-r: Great 👍 , thanks for this consistency

safydy-r (approved):
Thanks @bzabos , I think vmo-4322 is ready for QA once you merge this ;)

PR #172, opened 2021-07-29
VMO-4193 - working flow edit as modal

jcbashdown:
Here is the corresponding voto5 PR - https://bitbucket.org/voto/voto5/pull-requests/1931/vmo-4193-changes-needed-for-edit-flow-as


src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -32,9 +32,23 @@
...
                 {{ trans('flow-builder.new-flow') }}
               </router-link>
+              <button
+                class="btn btn-outline-primary btn-sm"
+                @click="showOrHideEditFlowModal">
+                {{ 'flow-builder.edit-flow' | trans }}
+              </button>
+              <b-modal
+                ref="edit-flow-modal"
+                title="Edit Flow"
```
poojahpatil90, line 45: Can you localize this please?

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -335,6 +353,28 @@ export default class TreeBuilderToolbar extends mixins(Routes, Permissions, Lang
...
+      }
+    }
+    this.$root.$on('bv::modal::show', () => {
+      this.$router.replace({
+        name: 'flow-details',
+      })
+    })
+    this.$root.$on('bv::modal::hide', () => {
+      this.$router.replace({
+        name: 'flow-canvas',
```
poojahpatil90, line 373: Thanks for hooking up the routes with the modal states 👍

poojahpatil90 (changes requested):
@jcbashdown Overall, looks good. Thanks for working on this and taking it off my plate 😃 
But I found a few issues:
1. The Blocks don't seem to be coming up due to an error. I noticed that this is not happening on master. Attached a screen recording to demo the issue. 
Bug1

2. The `Fix Issue` button in the notification bar does not open the modal. In the `ErrorNotifications.vue::fixFlowError()` method we are redirecting the user to `flow-details` route. But it does not seem to be working for some reason. Can you take a look at this?  Let me know if you need more context or help from my side.

bzabos (approved):
Hi @jcbashdown! I checked this out, looks good! 

I wonder if we should swap the Done button to the X in the top right? 🤷‍♀️  Either way, I like this better than the sidebar! Thanks!

poojahpatil90 (approved):
@jcbashdown Thanks for fixing those issues. Everything looks good to me. Approved!

bzabos (approved):
Hi @jcbashdown! Overall, this looks good and happy to merge 👌 

A concern of mine: I'd really like to discourage us from using watchers in vue. Most often these are a code smell and introduce challenges when trying to understand how the code _should_ work because it interrupts logical flows with background flows of "magic"— this also creates the conditions for challenges when debugging. 

Would you be willing to help me understand the ways in which changes to `$route.meta` would be unpredictable and unable to be architected explicitly? Thanks!

bzabos:
> A concern of mine: I'd really like to discourage us from using watchers in vue. Most often these are a code smell and introduce challenges when trying to understand how the code _should_ work because it interrupts logical flows with background flows of "magic"— this also creates the conditions for challenges when debugging.
> 
> Would you be willing to help me understand the ways in which changes to `$route.meta` would be unpredictable and unable to be architected explicitly? Thanks!

Following up on this from some slack DMs:

We can revisit this— I think we have some events as routes change, like: https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

```
  beforeRouteEnter(to, from, next) {
    // called before the route that renders this component is confirmed.
    // does NOT have access to `this` component instance,
    // because it has not been created yet when this guard is called!
  },
  beforeRouteUpdate(to, from, next) {
    // called when the route that renders this component has changed.
    // This component being reused (by using an explicit `key`) in the new route or not doesn't change anything.
    // For example, for a route with dynamic params `/foo/:id`, when we
    // navigate between `/foo/1` and `/foo/2`, the same `Foo` component instance
    // will be reused (unless you provided a `key` to `<router-view>`), and this hook will be called when that happens.
    // has access to `this` component instance.
  },
  beforeRouteLeave(to, from, next) {
    // called when the route that renders this component is about to
    // be navigated away from.
    // has access to `this` component instance.
  }
```

That was my gut's goto thought. Perhaps we could leverage routeEnter+Update events for handling these values. I wonder if @Pooja Patil’s other routing work might be able to leverage these events a bit as well.

PR #180, opened 2021-08-05
VMO-4122 - format imported json even if it is not a valid flow container

jcbashdown:


src/store/flow/views/import.ts
```diff
@@ -146,6 +146,7 @@ export const actions: ActionTree<IImportState, IRootState> = {
     try {
       // check valid json
       flowContainer = JSON.parse(state.flowJsonText)
+      commit('setFlowJsonText', JSON.stringify(flowContainer, null, 2))
```
jcbashdown, line 149: Once we know it's valid json we can set it formatted immediately. We then perform other validation steps.

safydy-r: @jcbashdown , for readability I'd prefer we put this after the `try catch`. Besides, that will avoid us confusion in case the error comes from `setFlowJsonText( )` or `JSON.stringify(flowContainer, null, 2)` (for now it's simple, but we may alter this mutation in the future)

What do you think?
I'm approving and let you decide which is the best.

safydy-r (approved):
left minor concern

src/store/flow/views/import.ts
jcbashdown: Good idea. I'll move it.

PR #181, opened 2021-08-06
Vmo 4464 fix flow builder block editor position

safydy-r:
**For tech:**
For some reason, the block editor position is affected by `interaction-designer-contents` element position. Therefore, as we have sidebar in voto5, we need to consider that when computing the block-editor's position.

**Voto5 PR** : https://bitbucket.org/voto/voto5/pull-requests/1934/fix-flow-builder-block-editor-position


jcbashdown (commented):
This doesn't seem to fix the issue for me? Not sure if I'm misunderstanding something...

jcbashdown (approved):
Looks good! Please make sure QA have the correct Jira ticket when testing.

PR #182, opened 2021-08-09
Vmo 4198 import page fixes

jcbashdown:
You may need to take a look at the associated DSN ticket to understand the requirements for this one as I'm experiencing some issues with the images in the ticket itself.


src/components/interaction-designer/flow-editors/import/ErrorHandler.vue
```diff
@@ -8,16 +8,26 @@
         <span class="sr-only">{{ 'flow-builder.loading' | trans }}</span>
       </div>
     </div>
+    <div
+      v-if="!flowError && !!flowJsonText"
+      class="alert alert-success mt-3"
+      role="alert">
+      <font-awesome-icon
+        :icon="['far', 'check-circle']"/>
```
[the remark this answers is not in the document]
safydy-r: A quick formatting here

src/views/ImportFlow.vue
```diff
@@ -285,5 +285,13 @@ export default ImportFlow
   overflow: hidden;
   white-space: nowrap;
 }
+</style>
+<style lang="scss" scoped>
+//The bootstrap default is what the design calls for but seems to get overridden somewhere else I can't find in order to style the block selection banner
+.alert-primary {
```
safydy-r, line 291: You're right about the `primary` colour we adapted it with Viamo colour but I don't know who made the change.
@jcbashdown , I'd suggest we use `alert-info` here instead of primary, that way we don't need to override the style.

safydy-r (approved):
Even I left a change suggestion for alert style to be `alert-info`, I'm approving this so you can move ahead after changes.

src/views/ImportFlow.vue
jcbashdown: It's a bit off from the design but I'm not sure Amen will mind and it's much cleaner. I'll make the change

PR #184, opened 2021-08-10
Vmo 4394 connection z index

poojahpatil90:
@bzabos A question for you - I notice that if we click on a connection and then try hovering over another, we don't have the desired behavior of highlighting the hovered line. I am wondering if we need to make a fix for this as well?  Or we can let the user uncheck the selection by clicking outside and then hover over the wires they want to see?


src/components/interaction-designer/Connection.vue
```diff
@@ -140,18 +141,16 @@ export default {
 
     this.line = new LeaderLine(start, end, this.options)
 
-    // Add event listeners
-    const self = this
```
poojahpatil90, line 144: I remove the reference to self as it did not make any difference in the functionality.  Let me know if there is concern around this change.

src/components/interaction-designer/Connection.vue
```diff
@@ -140,18 +141,16 @@ export default {
...
-    // the only way to identify current line so far: https://github.com/anseki/leader-line/issues/185
-    const connectionElement = document.querySelector('body>.leader-line:last-of-type')
+    this.connectionDomHandle = document.getElementById(`leader-line-${this.line._id}-line-path`).closest('svg.leader-line')
 
-    connectionElement.addEventListener('click', self.clickHandler, false)
+    // Add event listeners
+    this.connectionDomHandle.addEventListener('click', this.clickHandler, false)
 
-    connectionElement.addEventListener('click', self.clickAwayHandler(connectionElement), false)
+    this.connectionDomHandle.addEventListener('click', this.clickAwayHandler(this.connectionDomHandle), false)
```
bzabos, line 149: @poojahpatil90 `this.clickAwayHandler(…` doesn't seem to return anything, would we be able to replace:
```
this.connectionDomHandle.addEventListener('click', this.clickAwayHandler(this.connectionDomHandle), false)
```

with:

```
this.clickAwayHandler(this.connectionDomHandle)
```

? 😸

src/components/interaction-designer/Connection.vue
```diff
@@ -182,19 +181,22 @@ export default {
...
         this.deactivateConnection({connectionContext: this.connectionContext})
       }
+      this.connectionDomHandle.style.zIndex = 0
       this.$emit('lineMouseOut')
     },
     clickHandler() {
       this.isPermanentlyActive = true
       this.activateConnection({connectionContext: this.connectionContext})
       this.activateBlock({blockId: null})
+      this.connectionDomHandle.style.zIndex = 1
```
bzabos, line 199: Generally, it's good practice to use orders of magnitude on `zIndex` values so that it's easier to slide others in between. Eg. 1, 10, 100, etc. then someone can easily slide something in at 11 or 99 if they need to tune the app as complexity grows and more layers are added.

Does 10 push the connection line higher than desired?

bzabos (approved):
@poojahpatil90 this looks good! I left a couple minor comments that I'll leave up to your discretion. Feel welcome to merge if you're happy with what's there!

bzabos:
> @bzabos A question for you - I notice that if we click on a connection and then try hovering over another, we don't have the desired behavior of highlighting the hovered line. I am wondering if we need to make a fix for this as well? Or we can let the user uncheck the selection by clicking outside and then hover over the wires they want to see?

@poojahpatil90 Can you create a JIRA for this and touch base with @safydy on the details? I don't have a clear answer off the top of my head right now.

src/components/interaction-designer/Connection.vue
poojahpatil90: @bzabos Using zIndex with 10 as value brings out the connection to the top like in the screenshot. 
Screen Shot 2021-09-15 at 9 33 03 PM

src/components/interaction-designer/Connection.vue
poojahpatil90: Done!

PR #185, opened 2021-08-11
Vmo 4403 fix toolbar position with scrolling

safydy-r:
**Note for devs:**
- To make sure the toolbar keeps the position `sticky` and behaves like `fixed`, I used this solution: https://stackoverflow.com/a/57452861/3256489
- I took the opportunity to silent some `NavigationDuplicated` errors like we do with validations https://github.com/FLOIP/flow-builder/blob/VMO-4403-fix-flow-builder-block-editor-position/src/components/interaction-designer/toolbar/ErrorNotifications.vue#L133

**Screenshot**
vmo-4403

### DIFF
--- a/src/components/interaction-designer/BuilderCanvas.vue
+++ b/src/components/interaction-designer/BuilderCanvas.vue
@@ -187,7 +187,7 @@ export default class BuilderCanvas extends Vue {
       return this.windowWidth - this.widthAdjustment
     }
 
-    const xPosition: number = this.blockAtTheLowestPosition.ui_metadata.canvas_coordinates.x
+    const xPosition: number = this.blockAtTheFurthestRightPosition.ui_metadata.canvas_coordinates.x
     const scrollWidth = xPosition + this.blockWidth + MARGIN_WIDTH_CORRECTION
 
     if (scrollWidth < this.windowWidth) {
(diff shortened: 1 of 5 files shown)


src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -186,7 +187,7 @@ export default class BuilderCanvas extends Vue {
       return this.windowWidth - this.widthAdjustment
     }
 
-    const xPosition: number = this.blockAtTheLowestPosition.ui_metadata.canvas_coordinates.x
+    const xPosition: number = this.blockAtTheFurthestRightPosition.ui_metadata.canvas_coordinates.x
```
safydy-r, line 190: fixed wrong copy-paste from `canvasHeight()`

poojahpatil90: Can we use this opportunity to start the first block from the left-top corner or center. Right now it starts from the bottom-right corner. Suggesting this as it feels like this is the right spot to make the change.
If it is a time consuming change, then we can handle it later.

poojahpatil90 (commented):
@safydy-r  I tested this fix by checking out your branch. For some reason it does not seem to be working on my machine. Here is a screen capture to demonstrate the same.  Let me know if you feel I might be missing something.

src/components/interaction-designer/BuilderCanvas.vue
safydy-r: @poojahpatil90 , I'd like to handle that later in a different ticket.

poojahpatil90 (approved):
Thanks for fixing the issue. Looks good to me now. Approved!

PR #186, opened 2021-08-20
VMO-4136 Validation should delegate to individual block stores to each validate themselves Part 1

poojahpatil90:
@safydy-r  I took the opportunity with this task to move some of the helper functions in validation store to `validationHelper.ts` file. Please go through it and let me know your thoughts.

### DIFF
--- a/src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
+++ b/src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
@@ -51,7 +51,7 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
     }, {root: true})
   },
 
-  validate_block({rootGetters}, {block, schemaVersion}: {block: IBlock, schemaVersion: string}) {
+  validate({rootGetters}, {block, schemaVersion}: {block: IBlock, schemaVersion: string}) {
     return validate_community_block({block, schemaVersion})
   },
 }
--- a/src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
+++ b/src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
@@ -4,8 +4,8 @@ import {IBlock, IBlockExit} from '@floip/flow-runner'
 import {IdGeneratorUuidV4} from '@floip/flow-runner/dist/domain/IdGeneratorUuidV4'
 import {IPrintBlock} from '@floip/flow-runner/src/model/block/IPrintBlock'
 import {defaultsDeep} from 'lodash'
+import {validateCommunityBlock} from '@/store/validation/validationHelpers'
 import {IFlowsState} from '../index'
-import {validate_community_block} from '@/store/validation/validationHelpers'
 
 export const BLOCK_TYPE = 'ConsoleIO.Print'
 
@@ -52,7 +52,7 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
   },
 
   validate({rootGetters}, {block, schemaVersion}: {block: IBlock, schemaVersion: string}) {
-    return validate_community_block({block, schemaVersion})
+    return validateCommunityBlock({block, schemaVersion})
   },
 }
 
--- a/src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
+++ b/src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
@@ -5,6 +5,7 @@ import {IdGeneratorUuidV4} from '@floip/flow-runner/dist/domain/IdGeneratorUuidV
 import {IPrintBlock} from '@floip/flow-runner/src/model/block/IPrintBlock'
 import {defaultsDeep} from 'lodash'
 import {IFlowsState} from '../index'
+import {validate_community_block} from '@/store/validation/validationHelpers'
 
 export const BLOCK_TYPE = 'ConsoleIO.Print'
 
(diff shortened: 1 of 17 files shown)


src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
```diff
@@ -50,6 +51,9 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
     }, {root: true})
   },
 
+  validate_block({rootGetters}, {block, schemaVersion}: {block: IBlock, schemaVersion: string}) {
```
safydy-r, line 54: I think the `validate_block` action should be `async`, right ? (According to ACs)

src/store/validation/index.ts
```diff
@@ -77,18 +61,13 @@ export const mutations: MutationTree<IValidationState> = {
...
 
 export const actions: ActionTree<IValidationState, IRootState> = {
-  async validate_block({state, commit, rootGetters}, {block}: {block: IBlock}): Promise<IValidationStatus> {
-    const {uuid: blockId, type: blockType} = block
-    const blockTypeWithoutNameSpace = blockType.split('.')[blockType.split('.').length - 1]
-    const validate = getOrCreateBlockValidatorFor(blockTypeWithoutNameSpace, rootGetters['flow/activeFlowContainer'].specification_version)
-    const key = `block/${blockId}`
+  async validate_block({state, commit, rootGetters, dispatch}, {block}: {block: IBlock}): Promise<IValidationStatus> {
+    const schemaVersion = rootGetters['flow/activeFlowContainer'].specification_version
+    const status = await dispatch(`flow/${block.type}/validate_block`, {block, schemaVersion}, {root: true})
```
safydy-r, line 66: This confirms, `flow/${block.type}/validate_block` should be `async`

src/store/validation/validationHelpers.ts
```diff
@@ -0,0 +1,127 @@
+import {IBlock} from '@floip/flow-runner'
```
safydy-r, line 1: Thanks for moving these out, I'm think this will let us to have cleaner `validation` module 🙏

src/store/validation/validationHelpers.ts
```diff
@@ -0,0 +1,127 @@
...
+    // remove `blocks` property from IFlow schema to avoid double validations
+    flowJsonSchema.definitions.IFlow.additionalProperties = true
+    delete flowJsonSchema.definitions.IFlow.properties.blocks
+
+    validators.set(validationType, createDefaultJsonSchemaValidatorFactoryFor(flowJsonSchema, '#/definitions/IFlow'))
+  }
+  return validators.get(validationType)!
+}
+
+export function validate_community_block({block, schemaVersion}: {block: IBlock, schemaVersion: string}): IValidationStatus {
```
safydy-r, line 111: Let's keep method name cameCase in this helpers: `validateCommunityBlock`

Normally, even for methods under `validation` module, we could avoid snake_casing there (no need to updated them now for `validate_flow` etc, I'm just explaining).
To my understanding, we have snake_cased prefix (eg: `block_setName`, `resource_delete`, etc) if we have multiple vuex module (block, resources, flow, etc) under 01 name space (`flow`)
Screen Shot 2021-08-24 at 6 29 41 PM

That helps us to have better code readability, and gives us a clue where to see the method implementation if we have something like:
- dispatch(`flow/validate_flow`, ...)
- commit(`flow/resource_delete`, ...)

safydy-r (approved):
@poojahpatil90 , I'm approving so you can quickly move to QA once you finish addressing the minor concerns I left there.

Thanks

src/store/validation/validationHelpers.ts
safydy-r: And I suggest we use camelCase for `validateBlock` in each specific block type module as well.

src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
safydy-r: 2nd thought:
I'm wondering if we can rename this to `validate( )` only. In fact, it's already under the block-type module.
Then in validation module, when we call `validate_block()`, we will dispatch `dispatch('flow/${block.type}/validate'` instead of `dispatch('flow/${block.type}/validate_block'`. Which is less confusing for me.

poojahpatil90: I agree with naming this method validate as it is inside the respective block stores.

poojahpatil90: Regarding `async` for `validate` method, I feel it may not be necessary as we are awaiting the dispatched action to finish in our `validation/index.ts validate_block` method as seen below. We are not having an asynchronous operation inside the validate method itself. 
Let me know if I am missing something?   I would be happy to explain or provide more clarity on what I was thinking here.

src/store/validation/validationHelpers.ts
poojahpatil90: Your welcome :)

src/store/validation/validationHelpers.ts
poojahpatil90: I agree, done!

src/store/flow/block-types/ConsoleIO_PrintBlockStore.ts
safydy: Thanks for clarifying @poojahpatil90 , ... let's move it to QA then. I'm still a bit confused about the `async` but I think it's ok as long as it's working :D

PR #195, opened 2021-10-04
Vmo 4401 remove block editor offset

bulatgab:


bulatgab:
@safydy-r 
I've removed `interactionDesignerBoundingClientRect` from the `Block.vue` code.
It doesn't seem to be used anywhere else. Is it okay if I remove it from the Vuex store and from the `InteractionDesigner.vue` as well?

bulatgab:
> @bulatgab , `interactionDesignerBoundingClientRect` was introduced to fix toolbar position with scrolling (see original PR: #185). After, it seems we made several css fixes as well.
> 
> Could you confirm if, with your current change, **the toolbar position is good** when you move the block a farther right and bottom ? (check both implementation flow-builder community, and flow-builder viamo) I checked the community version, and it seems good, but pls double check.
> 
> If it does work, you can remove all `interactionDesignerBoundingClientRect` references from our code I think.
> 
> (See above PR to see a gif)

@safydy-r I have tested via the Viamo integration (voto5 + flow-builder) and it turns out the issue with the gap between the block and the block editor is back after my changes. Which means I will need to redo this PR.

PR #199, opened 2022-01-12
Vmo 4608 fix validation after import

jcbashdown:
Worth looking at without whitespace changes. 

This basically allows us to wait for all the block stores to be set up before initialising the active tree. Initialising the active tree triggers validations which are delegated to the block stores since this PR:

https://github.com/FLOIP/flow-builder/pull/186/files

Because these validations return undefined when the store isn't there (rather than erroring), the validation key for that block gets it's value set to undefined which the ErrorNotifications component cannot handle. This ensures we don't try anything until the stores are ready.


src/views/InteractionDesigner.vue
```diff
@@ -264,6 +268,7 @@ export default {
         invoke(blockClass, 'install', this)
         Vue.component(`Flow${typeWithoutSeparators}`, uiComponent)
       })
+      return Promise.all(blockInstallers)
```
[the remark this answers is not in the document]
jcbashdown: I think what I thought yesterday is basically correct - you can't wait for an async function within a forEach. See here:

https://advancedweb.hu/how-to-use-async-functions-with-array-foreach-in-javascript/

But you're right, we can have an await here instead of then so I've done that!

src/views/InteractionDesigner.vue
```diff
@@ -251,7 +253,8 @@ export default {
     async registerBlockTypes() {
       const {blockClasses} = this
 
-      forEach(blockClasses, async (blockClass) => {
+      const blockInstallers = keys(blockClasses).map(async (blockClassKey) => {
+        const blockClass = blockClasses[blockClassKey]
```
bulatgab, line 257: Minor suggestion. Would it work if we did it without keys? Like this:
```
const blockInstallers = blockClasses.map(async (blockClass) => {
   const type = blockClass.type
   ...
}
```

src/store/builder/f0492216-65f3-4dc7-a1d8-8f79d1d30c11-flow.json
```diff
@@ -100,6 +100,49 @@
...
+            },
+            {
+              "uuid": "9c7c4aa0-14ed-482f-b56d-02ebb0ada9c1",
+              "name": "Default",
+              "default": true,
+              "test": "",
+              "config": {}
+            }
+          ],
+          "tags": []
```
bulatgab, line 145: What is the purpose of these json files? Are they for testing?

jcbashdown: Yes, they're test files. Adding this block is intended to show that you can import a tree with blocks that aren't totally valid (`is_member` is null here). The intention is that the UI for fixing this kind of thing is well developed in the InteractionDesigner.

bulatgab (approved):
Approved.
Didn't test it myself though.

src/views/InteractionDesigner.vue
bulatgab: Ah, true

bulatgab: But that would save only one line at best, so I don't insist 😄

src/store/builder/f0492216-65f3-4dc7-a1d8-8f79d1d30c11-flow.json
bulatgab: Okay. Do you mean unit tests? Or manual tests? In any case, how can I try and see it in action?

jcbashdown: This is for manual testing. You can try it on the import page and see how it renders in the Interaction Designer.

src/views/InteractionDesigner.vue
jcbashdown: It's a good suggestion! Much more straightforward.

src/store/builder/f0492216-65f3-4dc7-a1d8-8f79d1d30c11-flow.json
bulatgab: Ah, I see now. I just had to use the `Paste JSON directly` option.
I thought that maybe the flow-builder was set up to read it automatically from the `src/store/builder` folder (along with other files) when in the dev mode 🙂

jcbashdown: You can also upload them - either way should work the same.

safydy-r (approved):
No more concerns from my end

jcbashdown:
**Notes for QA:**

To test this. First on master:
1) Go to the import page and import the valid flow in `src/store/builder/5b8c87d6-de90-4bc4-8668-4f040000006e-valid-flow.json`
2) You won't see any validation errors on the Interaction Designer page
3) You will see errors in the console

Switch to this branch, rebuild, refresh your browser and tree again:

1) You should see validation errors for the GroupBlock and the NumericBlock
2) There should be no console errors

You may also want to test that things still work as on master when you create a new flow (instead of importing). Also test that the Edit Flow modal on the Interaction Designer still works correctly and the links to individual validation errors work as they dp on master (you can test on master by creating a new flow instead of importing as validations aren't broken there)

pascaleli (approved):

PR #200, opened 2022-01-13
Vmo 4152 resource validation

safydy-r:
This PR implements main `resource validation`, except for:
- validation on expressions (https://viamoinc.atlassian.net/browse/VMO-5190)
- deep linking to select the right language tabs (https://viamoinc.atlassian.net/browse/VMO-5191)
- Update resources when we changes the activeFlow's languages (https://viamoinc.atlassian.net/browse/VMO-5213)

Merge plan:
- merge flow-runner PR first: https://github.com/FLOIP/flow-runner/pull/59
- update flow-builder package.json, replace `"@floip/flow-runner": "https://github.com/FLOIP/flow-runner.git#VMO-4152-resource-validation",` to `"@floip/flow-runner": "https://github.com/FLOIP/flow-runner.git#feature/VMO-4281-flow-spec-rc2`
- merge this PR

[Overview]
vmo-4152


package.json
```diff
@@ -23,7 +23,7 @@
   "dependencies": {
     "@avcs/autosuggest": "^1.7.5",
     "@floip/expression-evaluator": "^2.1.0",
-    "@floip/flow-runner": "https://github.com/FLOIP/flow-runner.git#feature/VMO-4281-flow-spec-rc2",
+    "@floip/flow-runner": "https://github.com/FLOIP/flow-runner.git#VMO-4152-resource-validation",
```
safydy-r, line 26: Will revert after merging https://github.com/FLOIP/flow-runner/pull/59

src/components/interaction-designer/resource-editors/ResourceVariantTextEditor.vue
```diff
@@ -1,107 +1,111 @@
 <template>
-  <div class="resource-variant-text-editor">
-    <div
-      class="content-editor"
-      :class="{'content-editor-selected': !!content}">
-
-      <div class="input-group">
-        <div v-if="label" class="input-group-prepend">
-          <span class="input-group-text">{{label}}</span>
+  <validation-message
```
safydy-r, line 2: Just wrapping everything within `<validation-message`

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
```diff
@@ -43,10 +43,12 @@ import {ErrorObject} from 'ajv'
 import {Prop} from 'vue-property-decorator'
 import {IFlow} from '@floip/flow-runner'
 import {namespace} from 'vuex-class'
+import {get, union} from "lodash"
 
 const flowVuexNamespace = namespace('flow')
+const validationVuexNamespace = namespace('validation')
 
-const DEFAULT_LIST_SIZE = 2
+const DEFAULT_LIST_SIZE = 3
```
safydy-r, line 51: @bulatgab , FYI here, I increased this as we have more validation errors to display

src/components/interaction-designer/toolbar/ErrorNotifications.vue
```diff
@@ -107,14 +107,7 @@ export default class ErrorNotifications extends mixins(Routes, Lang) {
   }
 
   get blockValidationStatuses(): { [key: string]: IValidationStatus } {
-    const blocksMap = this.activeFlow?.blocks.reduce((map: { [key: string]: boolean }, block) => {
```
safydy-r, line 110: When thinking about this block of code, I noticed we can remove the mapping because block's validation are already prefixed with `block/`

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
bulatgab: Sounds good!

bulatgab:
The jsdoc for `flattenErrorMessages()` (in `flow-builder/src/store/validation/index.ts`) has examples with flow and block errors. Could you add an example of a resource error to that list as well, please?

src/components/interaction-designer/resource-editors/ResourceEditor.vue
```diff
@@ -225,6 +231,16 @@ export class ResourceEditor extends mixins(FlowUploader, Permissions, Routes, La
...
 
+  /**
+   * Compute resource index (cell index) for a table having X languages and Y modes
+   *
+   * @param langIndex
+   * @param modeIndex
+   */
+  computeResourceIndex(langIndex: number, modeIndex: number): number {
+    return langIndex * this.flow.supported_modes.length + modeIndex
+  }
```
bulatgab, line 242: As we discussed, `computeResourceIndex` is currently fragile (if we change the number of modes for the flow).
It would be great to find a way to make it more robust.

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
```diff
@@ -43,10 +43,12 @@ import {ErrorObject} from 'ajv'
 import {Prop} from 'vue-property-decorator'
 import {IFlow} from '@floip/flow-runner'
 import {namespace} from 'vuex-class'
+import {get, union} from "lodash"
```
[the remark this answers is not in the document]
bulatgab: minor eslint issues like this can be fixed

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
```diff
@@ -57,12 +59,21 @@ export default class BlockErrorsExpandable extends mixins(Lang) {
...
+      : this.allErrors?.slice(0, DEFAULT_LIST_SIZE) ?? []
+  }
+
+  /**
+   * get allErrors to consider errors from:
+   * - block validation
+   * - resource validation
+   */
+  get allErrors() {
+    return union(this.status.ajvErrors, this.resourceValidationStatuses.ajvErrors);
```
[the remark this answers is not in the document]
bulatgab: minor eslint issues like this can be fixed

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
```diff
@@ -57,12 +59,21 @@ export default class BlockErrorsExpandable extends mixins(Lang) {
...
+      : this.allErrors?.slice(0, DEFAULT_LIST_SIZE) ?? []
+  }
+
+  /**
+   * get allErrors to consider errors from:
+   * - block validation
+   * - resource validation
+   */
+  get allErrors(): ErrorObject[] {
+    return union(this.status.ajvErrors, this.resourceValidationStatuses.ajvErrors);
```
bulatgab, line 72: The block validation errors are passed down from the parent component as a 'status' _prop_, whereas resource validation errors are read from the _vuex store_.
Do you think it would make sense to unify this and read the block validation errors from the store as well?

P.S. I think I was the one who did it via the props, ignoring the global store approach. My bad 😄

src/components/interaction-designer/resource-editors/ResourceEditor.vue
safydy-r: @bulatgab , thanks for raising flag here. I do agree using number `index` here could be confusing, but that's how the jsonSchema validate values of array (`resources.values` is an array).
I spent time to find alternative options, but couldn't so far.
Fortunately, I was able to correct the resources.values when users update `modes`; but it gets a bit tricky when they update languages. 

Therefore, I created a new ticket to handle that: https://viamoinc.atlassian.net/browse/VMO-5213

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
safydy-r: @bulatgab, I think we don't really need to always get data from vuex (or always send them from parent).
The key point here: As`BlockErrorsExpandable.vue` is handling errors for **01 block**. So, if I change `resourceValidationStatuses` into `resourceValidationStatusesForCurrentBlockValidationStatus`, that would be less confusing. What do you think ?

src/components/interaction-designer/flow-editors/FlowEditor.vue
```diff
@@ -78,19 +78,31 @@ class FlowEditor extends mixins(Lang) {
   @Prop({default: 'flow-builder.edit-flow'}) readonly flowHeader!: string
   @Prop({default: true}) readonly sidebar!: boolean
 
-  updateFlowLanguages(value: ILanguage[] | ILanguage): void {
+  async updateFlowLanguages(value: ILanguage[] | ILanguage): Promise<void> {
     this.flow_setLanguages({flowId: this.flow.uuid, value})
+    await this.validate_resourcesOnSupportedValues({
```
safydy-r, line 83: @bulatgab : I made update here after your 1st round review

src/components/interaction-designer/flow-editors/FlowEditor.vue
```diff
@@ -78,19 +78,31 @@ class FlowEditor extends mixins(Lang) {
...
+      resources: this.resourcesOnActiveFlow,
+      supportedModes: this.activeFlow.supported_modes
+    })
   }
 
-  updateFlowModes(value: SupportedMode[] | SupportedMode): void {
+  async updateFlowModes(value: SupportedMode[] | SupportedMode): Promise<void> {
     this.flow_setSupportedMode({flowId: this.flow.uuid, value})
+    await this.validate_resourcesOnSupportedValues({
```
safydy-r, line 91: @bulatgab : I made update here after your 1st round review

src/components/interaction-designer/flow-editors/LanguagesEditor.vue
```diff
@@ -35,7 +36,8 @@ class LanguagesEditor extends mixins(Lang) {
   @Prop() readonly flow!: IFlow
 
   get languages(): ILanguage[] {
-    return this.$store.state.trees.ui.languages
+    // Make sure to follow order when populating languages, because the order may affect indexes during resource validation
```
safydy-r, line 39: @bulatgab : I made update here after your 1st round review

src/store/flow/flow.ts
```diff
@@ -122,12 +122,15 @@ export const mutations: MutationTree<IFlowsState> = {
 
   flow_setSupportedMode(state, {flowId, value}) {
     const flow: IFlow = findFlowWith(flowId, state as unknown as IContext)
-    flow.supported_modes = Array.isArray(value) ? value : [value]
+    const supportedModesOrder = Object.values(SupportedMode)
+    // Make sure to follow order when populating supported_modes, because the order may affect indexes during resource validation
```
safydy-r, line 126: @bulatgab : I made update here after your 1st round review

src/store/flow/flow.ts
```diff
@@ -122,12 +122,15 @@ export const mutations: MutationTree<IFlowsState> = {
...
+    flow.supported_modes = Array.isArray(value) ? sortBy(value, (o) => supportedModesOrder.indexOf(o)) : [value]
   },
 
   flow_setLanguages(state, {flowId, value}) {
     const flow: IFlow = findFlowWith(flowId, state as unknown as IContext)
-    flow.languages = Array.isArray(value) ? value : [value]
+    // Make sure to follow order when populating languages, because the order may affect indexes during resource validation
+    flow.languages = Array.isArray(value) ? sortBy(value, ['label']) : [value]
```
safydy-r, line 133: @bulatgab : I made update here after your 1st round review

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
bulatgab: Okay, a mixed approach then.
Regarding the `resourceValidationStatuses` - personally, it wasn't confusing to me, but I'm okay with the new name.

src/components/interaction-designer/resource-editors/ResourceEditor.vue
jcbashdown: I also have some concerns here. It looks like you are making the index of resource language modes predictable by sorting when languages and modes change and then using the predictability to look up using an integer index. This is going to be fragile whatever I think - what if someone forgets to sort in future in a new place these values are changed? It also doesn't seem to work as intended though as the resource languages order doesn't seem to match this sorted order?

I understand this may be fixed in 5213. Something we might want to consider then is running a separate validation for values outside of the full resource validation. That way we can validate more specifically based on the content type (e.g. only allow a file path for certain content types) and we can also key the validation errors however we want rather than relying on the structure of the resource. For example, we can have string keys based on the resource uuid/mode/language id combination rather than relying on the order of an array. Does this make sense? Maybe this would be worse as it would involve a pretty different approach to what we use elsewhere but maybe also worth considering to remove the sort and resolve other issues here?

jcbashdown (changes requested):
Looks good. Just a couple of things that might need changing and I think there are some linting errors on lines you've changed to address as well. Then in the tickets you've spun out we might also want to do a refactor to make things more robust to changes languages (and modes).

src/components/interaction-designer/resource-editors/ResourceEditor.vue
safydy-r: @jcbashdown , thanks for input here.
My main blocker is that if we validate an array using AJV, the returned corresponding `dataPath` is `numeric indexed`, see this vuex state for eg:
Screen Shot 2022-01-25 at 3 49 22 PM

The numeric index corresponds to the order of the Data in the array, that's why it's important to follow a certain order. I know this is quite fragile, and I'm open if we can find alternative. The only solution I found so far:
- force order (like I did in this PR)
- change the resources spec from `array` to object

Any suggestion?

src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
```diff
@@ -72,16 +83,19 @@ export default class BlockErrorsExpandable extends mixins(Lang) {
       : this.trans('flow-builder.untitled-block')
   }
 
+  get resourceValidationStatusesForCurrentBlockValidationStatus(): IValidationStatus {
```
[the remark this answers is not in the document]
safydy-r: Thanks, this would be better I think. I changed it !

src/components/interaction-designer/toolbar/ErrorNotifications.vue
safydy-r: @jcbashdown , nice catch here, I think you're right !
I reverted by changes and added a doc string to the getter to make sure we understand that mapping part.

src/components/interaction-designer/resource-editors/ResourceEditor.vue
jcbashdown: The issue I was finding - the actual ordering of the language values in the resource on your branch are not consistent. They still depend on the order in which I add the language. That means I can add the second language in the list on creation and then go back and add the first. Now when I edit the second tab in the UI the first becomes valid and vice versa.

This issue isn't quite the same as the one in VMO-5213 as that's about the fact that existing resources don't get new empty values when you add a new language.

The solution I'm proposing for the issue I've found here is that we are probably going to need to validate resource value strings directly and build up our own nested structure of validation errors corresponding to the IDs of language and modes. This is much more robust to edits of languages and modes than what we have but sacrifices the convenience of using the json schema for resources. It has the advantage though, of potentially letting us create specific validators for each content type.

Let's try and have a call at some point to discuss further.

jcbashdown (approved):

jcbashdown:
I'm approving this pending 3 things:
1) Resolution of the conflict + merging in master before QA
2) Fixing of linting errors on new/edited lines
3) A quick spike on fixing the problems with validating (and displaying validation errors) when languages are added and removed. We agreed in a call to try and fix this by trying to add values to resources immediately when this happens and by sorting within resources.

If 3 proves complicated then we will do this in it's own ticket - VMO-5213

src/components/interaction-designer/resource-editors/ResourceEditor.vue
safydy-r: As discussed in our call, we will handle languages related behaviour in https://viamoinc.atlassian.net/browse/VMO-5213

PR #201, opened 2022-01-24
Vmo 4599 disable modes in voto5

bulatgab:
The main change is in the `ModesEditor` component. We now read `supportedModes` from the vuex store (where voto5 puts the updated/configured value - see voto5 PR), instead of reading it from the flow-runner dependency.

In `store/trees` we just add the `supportedModes` to the store.

The changes in `store/builder` are just refactoring.

In `store/flow` we change the `DEFAULT_MODES`. This variable is used for the initial selection of modes for newly created flows (the checkboxes) and imported flows (if not set). It's now limited to SMS, as that's the only mode in the list now.


safydy-r (approved):
Good job here @bulatgab , no concerns from my side

bulatgab:
> Good job here @bulatgab , no concerns from my side

Thank you for reviewing! @safydy-r

src/store/flow/flow.ts
```diff
@@ -401,6 +401,4 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
 
 export const DEFAULT_MODES = [
   SupportedMode.SMS,
```
jcbashdown, line 403: Hmm, I recognise this wasn't in the AC but wondering if this can also be configurable?

jcbashdown (changes requested):
Looks good! Left one comment a change that might be nice to have outside of the current AC. Do you think this would be quick to add?

src/store/flow/flow.ts
bulatgab: Thank you for the idea, I will add it

jcbashdown (approved):

jcbashdown (approved):
Thought of one other thing! Could you add a really quick description of how to use this in `docs/BUILDER_INTEGRATION.md` under the `Documentation of builder.config.json options` section? Approving pending this change.

PR #204, opened 2022-02-10
Vmo 5186 improve interaction designer customisation

bulatgab:


src/lib.ts
```diff
@@ -77,6 +79,7 @@ const Components: { [key: string]: any } = {
   NewFlowComponent,
```
bulatgab, line 79: Why are we globally registering other components under these names?
For example, why not `NewFlow: NewFlowComponent`?
(If the library users will be overriding them in the embedding apps, wouldn't they be registering them under the names they have in their Vue files, e.g. `NewFlow`?)

object-required (commented):
LGTM, but please keep in mind I'm not familiar with the project.

src/views/InteractionDesigner.vue
```diff
@@ -55,7 +55,7 @@ import {store} from '@/store'
 // import TreeViewer from './TreeViewer'
 // import LegacyInteractionDesigner from './InteractionDesigner.legacy'
 // import TreeUpdateConflictModal from './TreeUpdateConflictModal';
-import TreeBuilderToolbar from '@/components/interaction-designer/toolbar/TreeBuilderToolbar.vue'
+// import TreeBuilderToolbar from '@/components/interaction-designer/toolbar/TreeBuilderToolbar.vue'
```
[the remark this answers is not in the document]
bulatgab: Thanks @object-required!
I've cleaned up the commented-out lines now.
@safydy-r @jcbashdown Let me know if you'd like them back.

src/lib.ts
jcbashdown: Good point and now I can't think why. Let's make the change.

src/lib.ts
```diff
@@ -61,6 +62,7 @@ export const FetchFlow = FetchFlowComponent
 export const NewFlow = NewFlowComponent
 export const ImportFlow = ImportFlowComponent
 export const Home = HomeComponent
+export const TreeBuilderToolbar = TreeBuilderToolbarComponent
```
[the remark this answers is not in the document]
jcbashdown: Honestly, a lot of this comes through trial and error rather than following a specific pattern. Very happy to be corrected if there is a better approach which achieves the same goal. Here I think I just wanted the `XComponent` `const`s so I could use them twice. 

One this that might be good to add here is an index file of everything in the toolbar directory in the same way as commonComponents. Long term we want almost everything overridable - See VMO-442

jcbashdown (changes requested):
Looks good. Just a couple of small changes that would be nice to have.

src/views/InteractionDesigner.vue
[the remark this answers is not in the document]
jcbashdown: Good to remove them I reckon! I think they stuck around in case Brett had something in mind. Now we're back to properly working on this repo we can choose our own path.

safydy-r (approved):
Thanks for working on this, I'm approving but let's make sure to address what has been discussed from other reviewers ;)

src/lib.ts
```diff
@@ -16,11 +16,13 @@ import {
 import {FontAwesomeIcon} from '@fortawesome/vue-fontawesome'
 //To register components globally by default
 import * as commonComponents from '@/components/common'
+import * as toolbarComponents from '@/components/interaction-designer/toolbar'
```
jcbashdown, line 19: @bulatgab Does this work okay? I would have thought you'd need to add an index file?

jcbashdown (approved):

src/lib.ts
jcbashdown: Thanks for making that change! Approved.

PR #207, opened 2022-03-12
Vmo 5591 fix storybook

safydy-r:
This has fixes to bring storybook back again. It doesn't resolve errors in storyshoots for now though.

### DIFF
--- a/src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
+++ b/src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
@@ -71,6 +71,7 @@ import Lang from '@/lib/filters/lang'
 import {get} from 'lodash'
 import {mixins} from 'vue-class-component'
 import ExpressionInput from '@/components/common/ExpressionInput.vue'
+import {TextEditor, ValidationMessage} from '@/components/common/'
 
 const flowVuexNamespace = namespace('flow')
 
@@ -80,6 +81,8 @@ const EMPTY_STRING_EXPRESSION = ''
 @Component({
   components: {
     ExpressionInput,
+    ValidationMessage,
+    TextEditor,
   },
 })
 class ContactPropertyEditor extends mixins(Lang) {
--- a/src/css/stories/InteractionDesigner.css
+++ b/src/css/stories/InteractionDesigner.css
@@ -1,0 +1,496 @@
+/*
+* Helper functions for layout
+* Useful for storybook style
+*/
+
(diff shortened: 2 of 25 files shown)


src/css/stories/InteractionDesigner.css
```diff
@@ -0,0 +1,496 @@
+/*
+* Helper functions for layout
+* Useful for storybook style
```
safydy-r, line 3: I don't know why we lost this file in current master, we've been having it since the beginning though

object-required (approved):
LGTM 👍

src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
```diff
@@ -80,6 +81,8 @@ const EMPTY_STRING_EXPRESSION = ''
 @Component({
   components: {
     ExpressionInput,
+    ValidationMessage,
```
jcbashdown, line 84: @safydy-r These were removed as part of testing the changes you made in VMO-442. Components are now registered globally in `main.ts` and `lib.ts`. The effect of this will be to break component overriding I think - e.g. we will always used these local dependencies over what is registered with `Vue.component` - https://github.com/FLOIP/flow-builder/pull/208/files#diff-e59662795ab51f95e1bf2a44d30dd0debbc194a499adffd1bb6a197bd188d4b5R107

Really should have remembered this crucial part when reviewing VMO-442. What we need to do instead to bring these dependencies back in is something like the loop in that link in `/story-utils/storeSetup` or somewhere similarly global to all stories.

We also need to remove all `  components: {` everywhere. This should have happened in VMO-442 - was a crucial part and something that QA could have actually caught! Would have beem quite and awkward QA though as we would have to compare a branch where some things are overridden with one where they aren't

jcbashdown (changes requested):

src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
safydy-r: Thanks @jcbashdown for this input. As agreed in slack, feel free to implement your suggestion and let us know when you have a ready PR ;) 
Thanks in advance

yeboaasare (approved):

jcbashdown:
@yeboaasare Just so you are aware, we're now doing this fix here:

https://github.com/FLOIP/flow-builder/pull/210

Apologies for not closing this one sooner!

PR #208, opened 2022-03-14
Vmo-442 finalize component override

safydy-r:


jcbashdown (approved):
Looks good. Thanks for this! Will make life a bit easier for future work.

In terms of QA - I think we can skip this? We'll find problems as we go. I guess we could ask for regression testing but it seems like overkill. What do you think?

safydy-r:
@jcbashdown , I've amended commits' msg. Sorry it was wrong VMO
About QA: I agree to not QA this branch.

PR #213, opened 2022-03-22
Vmo 3121 tidy up builder embedding

bulatgab:


package.json
```diff
@@ -52,14 +52,15 @@
...
-    "leader-line": "^1.0.5",
+    "leader-line": "^1.0.7",
     "lodash": "^4.17.19",
     "m-class-list": "^1.1.9",
     "moment": "^2.27.0",
     "node-sass": "^4",
     "plain-draggable": "^2.5.12",
     "pointer-event": "^1.0.0",
     "popper.js": "^1.16.1",
+    "skeleton-loader": "^2.0.0",
```
safydy-r, line 63: Can you guys give a summary why this is a solution?

safydy-r (approved):
Approving, thanks @bulatgab & @jcbashdown

jcbashdown (approved):
Looks good. Thanks Bulat!

package.json
bulatgab: To my knowledge, the `leader-line` library is quite old and it doesn't support js modules, i.e. it doesn't have `export` in it. So we have to use a `skeleton-loader` to add it.

PR #215, opened 2022-04-01
VMO-5680 Expression validation ❗ Do NOT merge

object-required:
This PR attempts using a custom `format` specification to employing AJV's feature to validate FLOIP Expressions.

https://user-images.githubusercontent.com/3971957/161676060-0de099c7-f927-4db6-bcdb-c3ccf9132d78.mp4

### DIFF
--- a/src/store/flow/block.ts
+++ b/src/store/flow/block.ts
@@ -81,14 +81,21 @@ export const mutations: MutationTree<IFlowsState> = {
     Vue.delete(findBlockOnActiveFlowWith(blockId, state as unknown as IContext).config!, key)
   },
   block_updateConfigByPath(state, {blockId, path, value}: {blockId: string, path: string, value: object | string}) {
-    // todo: this might still break reactivity
-    // Make nested assignment reactive
-    setWith(
-      findBlockOnActiveFlowWith(blockId, state as unknown as IContext).config!,
-      path,
-      value,
-      (nestedValue, key, nestedObject) => Vue.set(nestedObject, key, nestedValue),
-    )
+    const base = findBlockOnActiveFlowWith(blockId, state as unknown as IContext).config
+    const chunks = path.split('.')
+
+    let pointer = base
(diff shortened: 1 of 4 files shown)


jcbashdown:
Nice! This along with the `flow-runner` change seems incredibly neat. Do you feel that anything is missing from this that will still take time/require thought in VMO-5090 or is this basically done?

jcbashdown:
Might be useful to link to the video in the ticket here also.

src/store/flow/block.ts
```diff
@@ -81,14 +81,21 @@ export const mutations: MutationTree<IFlowsState> = {
...
   block_updateConfigByPath(state, {blockId, path, value}: {blockId: string, path: string, value: object | string}) {
-    // todo: this might still break reactivity
-    // Make nested assignment reactive
-    setWith(
-      findBlockOnActiveFlowWith(blockId, state as unknown as IContext).config!,
-      path,
-      value,
-      (nestedValue, key, nestedObject) => Vue.set(nestedObject, key, nestedValue),
-    )
+    const base = findBlockOnActiveFlowWith(blockId, state as unknown as IContext).config
```
safydy-r, line 84: thanks for trying to address the TODO, wonder if this approach `cloneDeep` would be more clearer ? https://bitbucket.org/voto/voto5/src/55efeb5596050fba0c9c06c0333f64d7ecef38a9/resources/js/trees/stores/trees.js#lines-198

safydy-r (approved):
Approving as spike PR, left comments for vmo-5090 though

src/store/flow/block.ts
safydy-r: @jcbashdown , fyi: To answer your question here-above, another thing that I flagged in the other [PR](https://www.youtube.com/watch?v=ihrUIPfvTh8) is that we need to consider IResourceValue interface as well.

PR #217, opened 2022-04-07
Vmo 441 base block slots

bulatgab:


src/components/interaction-designer/block-types/Core_SetContactPropertyBlock.vue
```diff
@@ -7,22 +7,21 @@
...
+      <slot
+        slot="resource-editors"
+        name="resource-editors" />
+      <slot
+        slot="extras"
+        name="extras">
+        <contact-property-editor :block="block" />
+      </slot>
+      <slot name="vendor-extras" />
+      <slot
```
safydy-r, line 19: This is interesting @bulatgab , I never knew this before. This is the equivalent of
```
<template slot="branching">
        <slot name="branching" />
</template>
```
right ? Could you share me the related doc explaining this?

safydy-r (approved):
This is a neat PR, 💯 . 
Just left 01 question about the doc, I'd like to have a look in case there might be other thing I don't know yet. In fact, I couldn't find it inside the official doc https://vuejs.org/guide/components/slots.html#dynamic-slot-names

src/components/interaction-designer/block-types/Core_SetContactPropertyBlock.vue
bulatgab: @safydy-r Yes, exactly.
I saw people using it somewhere on forums, probably at vue github issues. But I couldn't find it now, where exactly it was, unfortunately.

bulatgab: @safydy-r 
Found it. I think it was this comment: https://forum.vuejs.org/t/set-grandchild-slot-content-via-parent/49995/3
See `<slot name="body" slot="body"></slot>`

src/components/interaction-designer/block-types/MobilePrimitives_SelectOneResponseBlock.vue
```diff
@@ -3,43 +3,40 @@
...
+          v-if="promptResource"
+          :label="'flow-builder.prompt' | trans"
+          :resource="promptResource"
+          :block="block"
+          :flow="flow" />
+      </slot>
+      <slot
+        slot="extras"
+        name="extras">
+        <choices-builder
```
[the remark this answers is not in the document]
bulatgab: @safydy-r I don't mind, the order of components would remain the same.
@jcbashdown What do you think?

src/components/interaction-designer/block-types/Core_SetContactPropertyBlock.vue
jcbashdown: This is great! Took me a while to get it. I myself copied a stack overflow approach without considering this much better one.

jcbashdown (approved):
I think this is basically good but see my comment about the extras slot. From the docs:

`    - extras - add content and UI specific to your custom block`

Let's emphasise to QA that there should be no functional changes and they should compare the UI on master to the UI on this branch carefully for each sidebar.

PR #219, opened 2022-04-11
Feature/vmo 5464 consume spec rc4

safydy-r:


package.json
```diff
@@ -23,7 +23,7 @@
   "dependencies": {
     "@avcs/autosuggest": "^1.7.5",
     "@floip/expression-evaluator": "^2.1.0",
-    "@floip/flow-runner": "1.0.0-rc2.7",
+    "@floip/flow-runner": "https://github.com/FLOIP/flow-runner.git#VMO-5464-initiate-rc4",
```
safydy-r, line 26: - [ ] Change this to `ˆ1.0.0-rc4` once flow-runner rc4 is merged

src/store/flow/block.ts
```diff
@@ -47,11 +47,6 @@ export const mutations: MutationTree<IFlowsState> = {
     const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
     findBlockExitWith(exitId, block).test = value
   },
-  block_setExitConfigByPath(state, {exitId, blockId, path, value}: {exitId: string, blockId: string, path: string, value: object | string}) {
```
safydy-r, line 50: Unused !

src/store/flow/resource.ts
```diff
@@ -28,48 +27,51 @@ import {IFlowsState} from '@/store/flow/index'
...
   resourceUuidsOnActiveFlow: (state, getters) => filter(map(getters.activeFlow.blocks, (block) => block.config.prompt)),
-
-  resourcesOnActiveFlow: (state, getters): IResource[] => filter(
-    state.resources,
-    (resource) => includes(getters.resourceUuidsOnActiveFlow, resource.uuid)
-  ),
 }
 
-export const mutations: MutationTree<IFlowsState> = {
-  resource_add({resources}, {resource}: { resource: IResource }) {
```
safydy-r, line 42: Moved these into actions, because we have access to `getters` there

src/store/flow/resource.ts
```diff
@@ -28,48 +27,55 @@ import {IFlowsState} from '@/store/flow/index'
...
+  resourceUuidsOnActiveFlow: (_state, getters) => filter(map(getters.activeFlow.resources, (res) => res.uuid)),
 }
 
-export const mutations: MutationTree<IFlowsState> = {
-  resource_add({resources}, {resource}: { resource: IResource }) {
-    resources.push(resource)
+export const mutations: MutationTree<IFlowsState> = {}
+
+export const actions: ActionTree<IFlowsState, IRootState> = {
+  resource_add({getters}, {resource}: {resource: IResource}) {
```
jcbashdown, line 38: What's the reason to move to an action here? I see you are catching a null resource, what was causing this situation?

Are there consequences to this now being async?

src/store/flow/resource.ts
```diff
@@ -28,48 +27,55 @@ import {IFlowsState} from '@/store/flow/index'
...
+      throw new ValidationException('Unable to add null resource to flow')
+    }
+
+    getters.activeFlow.resources.push(resource)
   },
 
   // currently unused - see todo
-  resource_delete({resources}, {resourceId}: { resourceId: string }) {
+  resource_delete({getters}, {resourceId}: {resourceId: IResource['uuid']}) {
+    const flow: IFlow = getters.activeFlow
```
jcbashdown, line 48: Again, using a getter seems weird to me. It's not how things are ever done in the docs and I'm concerned it might have unintended consequences. From the docs:

> Actions are similar to mutations, the differences being that:
>
> Instead of mutating the state, actions commit mutations.
> Actions can contain arbitrary asynchronous operations.

So that suggests that actions should not change the state directly and then the fact that mutations don't have access to getters is suggestive of this being bad practice.

See this stackoverflow answer:

https://stackoverflow.com/questions/43258793/accessing-getters-within-vuex-mutations

jcbashdown (changes requested):
Looks mostly good. Just a few questions and change suggestions

src/store/flow/resource.ts
jcbashdown: See my comment about this.

safydy-r: OK 👁️

src/components/common/AudioLibrarySelector.vue
```diff
@@ -27,7 +27,7 @@
 <script lang="js">
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types,@typescript-eslint/strict-boolean-expressions */
 import {isEmpty} from 'lodash'
-import {mapActions} from 'vuex'
+import {mapActions, mapGetters} from 'vuex'
```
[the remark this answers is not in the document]
safydy-r: I tested the `getters.activeFlow` in this file, and forgot to remove this when I removed the code. Getting this out. Thanks

src/store/builder/index.ts
```diff
@@ -299,30 +299,33 @@ export const actions: ActionTree<IBuilderState, IRootState> = {
...
    * @param dispatch
    * @param commit
    * @param state
    * @param rootState
    * @param flows
    */
-  async importFlowsAndResources({rootState}, {flows, resources}: { flows: IFlow[], resources: IResource[] }) {
-    console.debug('importing flows & resources ...')
-    console.log({flows, resources})
+  async importFlowsAndResources({rootState}, {flows}: { flows: IFlow[]}) {
```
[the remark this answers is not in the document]
safydy-r: @jcbashdown , I think it's not even used by the exporter. In fact, we don't need a setter from that textarea which has the flow string. I took the opportunity to:
- remove this setter https://github.com/FLOIP/flow-builder/blob/master/src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue#L396,
- turn v-model to a simple value here https://github.com/FLOIP/flow-builder/blob/master/src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue#L11
- remove this action `importFlowsAndResources()`

See the commit `b1d5812` for other associated changes

src/store/builder/2019-10-10-shortcut-flow.json
```diff
@@ -1,15 +1,38 @@
 {
+  "specification_version": "1.0.0-rc4",
```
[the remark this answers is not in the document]
safydy-r: Updated !
@jcbashdown , could you pls remind me what's the purpose of these json fixtures in flow-builder ?
For me, I just use them to manually test the importer feature. I don't know if the was an original purpose behind them.

src/store/flow/views/import.ts
```diff
@@ -66,16 +66,16 @@ export const mutations: MutationTree<IImportState> = {
   setExistingLanguagesWithoutMatch(state, languages) {
     state.existingLanguagesWithoutMatch = languages
   },
-  setFlowLanguages(state, languages) {
+  setFirstFlowLanguages(state, languages) {
```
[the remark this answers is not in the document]
safydy-r: Ok, I renamed & added comment for the 02 mutations

src/store/flow/resource.ts
safydy-r: The main reason to use `actions` are to access at the `getters.activeFlow` which avoids us to spread `flowId` as param everywhere.

After seeing your comment below, I brought changes on these codes. Pls review again.

src/store/flow/resource.ts
safydy-r: @jcbashdown , I agree on this! thanks for pointing this out. After spending hours on making flow-builder working again, I forgot this notion. 

Pls review again the code.

src/store/flow/resource.ts
```diff
@@ -28,48 +27,55 @@ import {IFlowsState} from '@/store/flow/index'
...
-    resources.push(resource)
+export const mutations: MutationTree<IFlowsState> = {}
+
+export const actions: ActionTree<IFlowsState, IRootState> = {
+  resource_add({getters}, {resource}: {resource: IResource}) {
+    if (resource == null) {
+      throw new ValidationException('Unable to add null resource to flow')
+    }
+
+    getters.activeFlow.resources.push(resource)
```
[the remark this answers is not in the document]
safydy-r: Agree, see my comment in your next comment

src/components/common/AudioLibrarySelector.vue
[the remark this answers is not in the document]
jcbashdown: :+1:

src/store/builder/index.ts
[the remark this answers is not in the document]
jcbashdown: Great! Thanks for making this change

src/store/builder/2019-10-10-shortcut-flow.json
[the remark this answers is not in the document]
jcbashdown: Same for me - I use them to test the importer and the various importer specific ways you can be prompted to resolve validation errors. It's possible that the one testing unmatched contact properties and groups is out of date but this will probably become relevant again when we reintroduce contact and group dropdowns in those blocks' uis.

src/store/flow/views/import.ts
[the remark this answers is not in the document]
jcbashdown: Thanks!

src/store/flow/resource.ts
jcbashdown: Thanks for making these changes! All looks good now.

jcbashdown (approved):

jcbashdown:
I think there might be a couple of linting errors and some conflicts but nothing major. All looks good to me. Approved!

PR #222, opened 2022-04-16
VMO-3996 User-friendly error messages

object-required:


jcbashdown (approved):
LGTM!

safydy-r (approved):
No comment here, thanks

adeablay:
@object-required Here are my findings on the other areas that have non-user friendly validation messages.
1. No matter the number of blocks that exists on the Flow canvas, user still gets a validation error message to add a block when the very first block that was added is deleted. This error however disappears when one of the blocks is set as the starting block. 
My findings: In the Viamo trees, when the first block is deleted, the next block is automatically set as the starting block. 



https://user-images.githubusercontent.com/79596952/164461266-dbdbdaa7-2f79-4f9c-80a6-1d150a145065.mov

adeablay:
2.  /first_block_id does not seem user friendly. My suggestion will be to take it off completely so we just have, _"Flow must contain at least one block"_  and _Add at least one block_  Whatever works will be fine I guess.
The text in red (Flow must contain at least one block) and text in black(Add at least one block) seem to be together on the same line. This makes reading it tricky at first glance.
add a block

adeablay:
3. Similar to my comment no. 2, /name, /value/0/value and the regular expressions also do not seem to be user friendly. Some of these may be related to VMO-5731. The image attached is for the message block. For VMO-5731, you can find similar instances in all other blocks especially the Content blocks, Set Group Membership and Output blocks.

image

PR #223, opened 2022-04-19
Vmo 3685 numeric block editor options

bulatgab:
This PR fixes the `vendor-extras` slot. Details reposted from Slack:

I’m working on VMO-3685 and this slot (that I added in VMO-441) doesn’t work, i.e. if I add these block editors
```
      <template slot="vendor-extras">        
        <block-repeat-questions-config :block="block"/>
        <block-response-timeout-config :block="block"/>
      </template>
```
in `Viamo_MobilePrimitives_NumericResponseBlock` they are not displayed.
This is happening because in `MobilePrimitives_NumericResponseBlock` I added `<slot name="vendor-extras" />` as a child of a `<base-block/>`. This lead to it being inserted into the _default_ slot of a BaseBlock which doesn’t exist (we have other named slots in BaseBlock, but not the default one).

Long story short, to solve it, the `vendor-extras` slot needs to be added to the `BaseBlock` and then re-exposed in `MobilePrimitives_NumericResponseBlock` and other blocks, the same way we do with `extras`, `resource-editors`, etc., rather than just being created in those blocks, as we do e.g. with the `vendor` slot.


safydy-r (approved):
Thanks for the explanation. I saw you & @jcbashdown had a long discussion about it. I'm trusting your decision here. Approved !

jcbashdown (approved):
Thanks for this change @bulatgab

safydy-r:
@bulatgab , one thing I noticed on this change: The sensitive data checkbox gets updated when we click on the section label (or on the tooltip)
Screen Shot 2022-05-12 at 7 26 24 PM

- [ ] let's avoid that, like with other sections (eg: Set Contact Prop)

bulatgab:
> @bulatgab , one thing I noticed on this change: The sensitive data checkbox gets updated when we click on the section label (or on the tooltip) Screen Shot 2022-05-12 at 7 26 24 PM
> 
> * [ ]  let's avoid that, like with other sections (eg: Set Contact Prop)

Thanks @safydy-r for reporting this.
This issue is specific to the changes in `voto5`.
The reason this was happening was because I used `<label>` for the 'Sensitive data' title as we do everywhere else. But I used with `for` attribute as required by the ESLint rules in `voto5`.
It is now replaced with `<p>` element with additional styling (`font-weight-bold` class) and without `for`.

PR #224, opened 2022-04-19
VMO-5765 fix b-tooltip errors on storyshots

safydy-r:
This has intention to fix 
```
console.error node_modules/vue/dist/vue.common.dev.js:630
    [Vue warn]: Failed to resolve directive: b-tooltip
    
    (found in <TreeBuilderToolbar>)
```

like in https://app.travis-ci.com/github/FLOIP/flow-builder/builds/249587733#L452

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -12,13 +12,9 @@ import 'vue-multiselect/dist/vue-multiselect.min.css'
 import 'scss/main.scss'
 
 import registerCustomComponents from '@/common-imports'
-import {BootstrapVue, BTooltip} from 'bootstrap-vue'
 
 import router from './router'
 import App from './App.vue'
-
-Vue.use(BootstrapVue)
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -12,14 +12,10 @@ import 'vue-multiselect/dist/vue-multiselect.min.css'
 import 'scss/main.scss'
 
 import registerCustomComponents from '@/common-imports'
-import {BootstrapVue, BTooltip} from 'bootstrap-vue'
 
 import router from './router'
 import App from './App.vue'
 
-Vue.use(BootstrapVue)
```
safydy-r, line 20: moving this back to vue component fixes the issue, so let's stick with that for now !

bulatgab: Sounds good. We can get back to this (to how we register bootstrap vue) later :)

PR #225, opened 2022-04-20
VMO-5818 make usesDefaultContactPropsEditor true by default for block…

safydy-r:
Putting back the component for these all MobilePrimitives & SmartDevices blocks.


bulatgab (approved):

bulatgab:
Thanks @safydy-r for the quick fix for all these blocks 👍

PR #226, opened 2022-04-21
Feature/vmo 5627 add list on set contact prop component

safydy-r:
[Overview]
vmo-5627-flow-builder


builder.config.json
```diff
@@ -287,7 +303,16 @@
     "treeTitles": {},
     "treeSetTitles": {},
     "groupNames": {},
-    "groups": [],
+    "groups": [
```
safydy-r, line 306: I populated `groups` to make the importer working (see gifs in the description)

src/components/interaction-designer/block-editors/GenericContactPropertyEditor.vue
```diff
@@ -195,13 +215,32 @@ export class GenericContactPropertyEditor extends mixins(Lang) {
...
+
   @flowVuexNamespace.Mutation block_updateConfigByPath!: (
     {blockId, path, value}: { blockId: string, path: string, value: string | object }
   ) => void
   @flowVuexNamespace.Mutation block_updateVendorMetadataByPath!: (
     {blockId, path, value}: { blockId: string, path: string, value: string }
   ) => void
   @flowVuexNamespace.Mutation block_removeConfigByKey!: ({blockId, key}: { blockId: string, key: string}) => void
+
+  @State(({trees: {ui}}) => ui) ui!: any
```
jcbashdown, line 243: Generally not a fan of this pattern. I would prefer using the subscriber property fields getter already in `src/store/trees/trees.js`

jcbashdown (changes requested):
Just one small change needed I think

src/components/interaction-designer/block-editors/GenericContactPropertyEditor.vue
safydy-r: Good point @jcbashdown , I think we already have that `subscriberPropertyFields` as getter there. I just updated the current component. Thanks 🙏

jcbashdown (approved):

src/components/interaction-designer/block-editors/GenericContactPropertyEditor.vue
```diff
@@ -27,7 +27,21 @@
         #input-control="{ isValid }"
         :message-key="`block/${block.uuid}/config/set_contact_property/property_key`">
         <div class="block-contact-property-key">
+          <div v-if="subscriberPropertyFields">
```
jcbashdown: @safydy-r This needs to be checking that this is empty rather than falsy. We could have the getter default to returning null but that would break the importer I believe and it's better to have a consistently typed getter. 

I've pushed a commit to fix this and bumped voto5 so I can move this to QA. Feel free to tweak my approach though.

yeboaasare (approved):

PR #231, opened 2022-04-27
feature/VMO 5605 fix the horizontal scroll on th

laserpants:
- fixed header
- PR-230 Remove redundant comma
- PR-230 Change noselect to no-select
- PR-230 Set top margin to avoid blocks being positioned outside of canvas


laserpants:
@safydy-r Please see my message on Slack. (I still don't have permissions to add reviewers.)

PR #221, opened 2022-04-14
VMO-5090 Validate expressions

object-required:

PR #232, opened 2022-04-27
VMO-5731 Validation messages for Expressions

object-required:
🛑 **Don't merge** this PR before #221 
The actual changes are in 558493f67c87f88dd255ccb43fe1b0018acb6c12


src/store/validation/validationHelpers.ts
```diff
@@ -89,13 +89,27 @@ export function debugValidationStatus(status: IValidationStatus, customMessage:
...
 }
 
-function getErrorMessageLocalizationKey(keyPrefix: string, ajvErrorObject: ErrorObject) : string {
+function getErrorMessageLocalizationKeyForProperty(keyPrefix: string, ajvErrorObject: ErrorObject) : string {
   const entity = keyPrefix.startsWith('flow') ? 'flows' : 'blocks'
-  const property = ajvErrorObject.dataPath.replaceAll('/', '-')
+  const property = ajvErrorObject.dataPath
+    .replaceAll('/', '-')
+    // Replacing digits with zeros to eliminate resource indexes
+    .replaceAll(/\d/g, '0')
```
safydy-r, line 97: let's replace them with `x` not `0`, to have something like `'base.x-times'` localization key

src/assets/messages.json
```diff
@@ -3363,7 +3366,8 @@
     "blocks-config-accuracy_threshold_meters-type": "Le seuil de précision en mètres devrait être un nombre",
     "blocks-config-accuracy_timeout_seconds-type": "Le délai d'attente de précision en quelques secondes devrait être un nombre",
     "blocks-config-value-type": "L'expression de sortie devrait être une chaîne",
-    "blocks-config-is_member-type": "La valeur / l'expression devrait être une chaîne"
+    "blocks-config-is_member-type": "La valeur / l'expression devrait être une chaîne",
+    "blocks-values-0-value-pattern": "Le contenu ne doit contenir que le script, les chiffres et la ponctuation de base"
```
safydy-r, line 3370: I know this was from the `getErrorMessageLocalizationKeyForProperty` logic. Is there a way to update that method so we will have `resources` word in this localization key ? Kind of `blocks-resources-values-x-value-pattern`

safydy-r (approved):
Good work here ! I'm pre-approving as well, to not block you after you address my inputs.

FYI: I updated the target branch to have a clean diff in this PR. Therefore, pls make sure you edit it to `master` before merge.

pascaleli:
@object-required  

This ticket will be sent to Blocked as well pending response from this comment 

https://github.com/FLOIP/flow-builder/pull/221#issuecomment-1113543761

pascaleli:
@object-required 

Can you please clarify this for me ?
Does the scope of this ticket include validation messages for resources such as choices in [select One and select many response block ?]
Because the ticket title made mention of "Validation message for resources & Expression" 

Users should at at least configure some choices like we have validation for this in the tree_builder.

image

PR #237, opened 2022-05-09
VMO-5605 Fix the horizontal scroll on the builder canvas in voto5

laserpants:
I found a workaround for `voto5` embedded mode. 

However, to get the flow-builder to work smoothly in both UIs will require more work, in particular: 
- a clear understanding of what the desired end-goal is, in terms of positioning, layout, and scroll behavior, e.g., what happens when blocks are moved outside of the visible canvas; 
- more experimentation and UI prototyping, which involves changes in both applications.


src/views/InteractionDesigner.vue
```diff
@@ -302,23 +302,29 @@ export default {
 <style lang="scss">
   $sidebar-width: 365px;
 
+html {
+  background: #F5F5F5;
+}
```
object-required, line 307: A minor indentation thing

object-required (approved):
I thought of using `flex` to allow the embedded component to resize to whatever dimensions are available. That didn't go exactly well 😅 but do you think it may be a feasible option?

markuphero-zBVIKhHbdavxCh1FBXAa (1)

safydy-r:
@laserpants , thanks for trying to find a solution here.

About your question above: As long as both community version & voto5 work, I'm ok to move extra refactoring in any tickets. But for your info, I tested the voto5 PR and it doesn't resolve the main problem. I left a screen recording there for details. Let me know if you have questions.


I'll wait for voto5 to be working before approving this PR, because you might bring new changes here.
- [ ] fix voto5 too

safydy-r:
@laserpants , thanks for trying to find a solution here.

About your question above: As long as both community version & voto5 work, I'm ok to move extra refactoring in any tickets. But for your info, I tested the voto5 PR and it doesn't resolve the main problem. I left a screen recording there for details. Let me know if you have questions.


I'll wait for voto5 to be working before approving this PR, because you might bring new changes here.
- [ ] fix voto5 too

laserpants:
Superseded by https://github.com/FLOIP/flow-builder/pull/250.

PR #239, opened 2022-05-11
Feature/vmo 5628/add custom validations for numeric block

safydy-r:
[overview]
vmo-5628-custom validation on numeric block

**Need confirmation from** @jcbashdown : 
the maxDigit hint is the same on Design (as tooltip) & in current community version (as small text under the input).
> Set maximum value will always override maximum response digit except when maximum value is empty

So, I think we don't need a voto5 customization in this ticket, we just need to transform the text into tooltip (according to the design). But pls confirm in case I missed something.


src/assets/messages.json
```diff
@@ -3351,7 +3351,12 @@
     "block-config-accuracy_timeout_seconds-type": "Accuracy timeout in seconds should be a number",
     "block-config-value-type": "Output expression should be a string",
     "block-config-is_member-type": "Value/Expression should be a string",
-    "resource-values-x-value-pattern": "Content should only contain basic Latin script, digits, or punctuation"
+    "resource-values-x-value-pattern": "Content should only contain basic Latin script, digits, or punctuation",
+    "numeric-block-min-value-must-be-lower-than-max-value": "Minimum value (inclusive) must be lower than Maximum value (inclusive)",
+    "numeric-block-max-value-must-be-greater-than-min-value": "Maximum value (inclusive) must be greatter than Minimum value (inclusive)",
```
object-required, line 3356:
```suggestion
    "numeric-block-max-value-must-be-greater-than-min-value": "Maximum value (inclusive) must be greater than Minimum value (inclusive)",
```

object-required: Perhaps, _Maximum value must be greater or equal to Minimum value_ would be easier to understand?

jcbashdown:
> [overview] vmo-5628-custom validation on numeric block
> 
> **Need confirmation from** @jcbashdown : the maxDigit hint is the same on Design (as tooltip) & in current community version (as small text under the input).
> 
> > Set maximum value will always override maximum response digit except when maximum value is empty
> 
> So, I think we don't need a voto5 customization in this ticket, we just need to transform the text into tooltip (according to the design). But pls confirm in case I missed something.

You're right. I must have missed that this wasn't changing when reviewing the designs. No need for voto5 customisation here.

src/assets/messages.json
safydy-r: @object-required , thanks for suggesting an improvement here. I added the `or equal` for this, and for the min-value above ! But for the rest, I'd like to leave it for designers, they will still revisit all those texts.

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -53,7 +53,7 @@
                 ref="edit-flow-modal"
                 ok-only
                 :ok-title="'flow-builder.done' | trans"
-                @ok="showOrHideEditFlowModal">
+                @ok="handleUpdateFlowDetails">
```
jcbashdown, line 56: The flow is updated whether you click okay to close the modal or you just click outside the modal which means that this doesn't quite work - I can add a new mode and it won't validate all resource fields

src/components/interaction-designer/block-types/MobilePrimitives_NumericResponseBlock.vue
```diff
@@ -75,12 +75,10 @@ class MobilePrimitives_NumericResponseBlock extends mixins(Lang) {
 
   updateValidationMin(value: number | string): void {
     this.setValidationMinimum({blockId: this.block.uuid, value})
-    this.handleActionsAccordingToBranchingType()
```
jcbashdown, line 78: Now that this is gone, is anything calling the function?

jcbashdown (changes requested):

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
safydy-r: Good catch @jcbashdown !
After thinking about this, I'm persuaded we should not have the close button for now & should not allow users to close the modal clicking outside. In fact, we bind the active flow itself to the form.

Therefore, see the associated commit https://github.com/FLOIP/flow-builder/pull/239/commits/1311ddd77fafe083131c6bfb95c79d1bbe81c840

Let me know if you have any concerns.

jcbashdown: Hmm. Can you explain a bit more about what was wrong with the old approach to this? Why not hook into changes on those fields?

The approach you are taking could have some advantages but it's a significant enough change to the UX that you might need to clear this with design.

jcbashdown (approved):

pascaleli:
@safydy-r 
Validation has been implemented for Maximum value can not be greater than the Maximum Response digits, in addition the Minimum  
The Ac says : Min value cannot be greater than the result of max digits. 

image


I get this console warning in the browser, when Maximum value is greater than Maxi Response 

`[Vue warn]: Duplicate keys detected: 'block/a0fa448d-aa5a-4cbd-9263-5721c79250bd/config/ivr/max_digits'. This may cause an update error.

found in

---> <BlockErrorsExpandable> at src/components/interaction-designer/toolbar/BlockErrorsExpandable.vue
       <ErrorNotifications> at src/components/interaction-designer/toolbar/ErrorNotifications.vue
         <TreeBuilderToolbar> at src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
           <Routes> at src/views/InteractionDesigner.vue
             <App> at src/App.vue
               <Root> `

PR #241, opened 2022-05-13
VMO-5825 Validation error for adding a block still shows even when there are existing blocks in the flow

object-required:
https://user-images.githubusercontent.com/3971957/168307964-346aa139-a426-4203-b404-a6c2cac7dfc6.mov


bulatgab (approved):
This changes the app's behaviour a little bit - now if the first block is deleted, the second block is reassigned _automatically_ without the user knowing it, unless they notice the new down-arrow on the new block.
I think it's not super-critical, it's even elegant in its simplicity.
I would suggest checking this with the designers though - maybe it would be better to show an additional notification about it (that the first block has changed). It shoudn't be too hard to implement and it will probably happen also here in this `if` statement. Can be done in a separate ticket.

PR #242, opened 2022-05-14
Feature/vmo 5540 validate viamo config for numeric block

safydy-r:
Main changes:
- to export helpers
- to tweak the validation action


object-required (approved):

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
```diff
@@ -68,21 +69,45 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
   },
 
+  /**
+   * Override this in the consumer side to add extra config props to avoid the validation saying we have missing prop at the creation
+   * eg: {
+   *   prop1: undefined,
+   *   prop2: undefined,
+   * }
+   */
+  async initiateExtraVendorConfig(_ctx: unknown): Promise<object> {
```
bulatgab, line 86: Is this function asynchronous in case users need it to be such (right now we don't require it to be)?

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
```diff
@@ -68,21 +69,45 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
+   *
+   * Important: This will be overridden in the consumer side, so DO NOT add generic validations here,
+   * instead edit the `validate()` if needed.
+   */
+  async validateVendorBlock({rootGetters}, {block, schemaVersion}: {block: IBlock, schemaVersion: string}): Promise<IValidationStatus> {
+    const blockTypeWithoutNameSpace = block.type.split('.')[block.type.split('.').length - 1]
+    return validateCommunityBlock({block, schemaVersion, customBlockJsonSchema: require(`@floip/flow-runner/dist/resources/validationSchema/${schemaVersion}/I${blockTypeWithoutNameSpace}Block.json`)})
+  },
+
+  async validate({rootGetters, dispatch}, {block, schemaVersion}: {block: IBlock, schemaVersion: string}) {
```
bulatgab, line 109: I wish the `validate` was called differently, to distinguish from `validateVendorBlock` and `validateCommunityBlock`.
Something like `validateAndUpdateErrorNotifications`(?). But that would mean changing this across all block stores.
Any ideas?

bulatgab (approved):

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
```diff
@@ -68,21 +69,45 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
+  /**
+   * Validate the vendor block (Consumer block)
+   * By overriding this action in the consumer side, we will be able to customize it using different json schema for eg.
+   *
+   * Important: This will be overridden in the consumer side, so DO NOT add generic validations here,
+   * instead edit the `validate()` if needed.
+   */
+  async validateVendorBlock({rootGetters}, {block, schemaVersion}: {block: IBlock, schemaVersion: string}): Promise<IValidationStatus> {
+    const blockTypeWithoutNameSpace = block.type.split('.')[block.type.split('.').length - 1]
+    return validateCommunityBlock({block, schemaVersion, customBlockJsonSchema: require(`@floip/flow-runner/dist/resources/validationSchema/${schemaVersion}/I${blockTypeWithoutNameSpace}Block.json`)})
```
jcbashdown, line 106: I think that validateCommunityBlock and validateVendorBlock are maybe no longer the right names. Here is a suggestion:

- `validateVendorBlock` -> `fetchBlockSchemaAndValidate`
- `validateCommunityBlock` becomes something like `validateBlockWithSchema`

Then I think you still only need to override the method now called `fetchBlockSchemaAndValidate` with vendor specific behaviour.

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
safydy-r: I'd keep it this short if you don't have a very strong opinion here. To reduce the confusing, let me add a js doc for that generic validate action. Thanks for bringing this up

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
safydy-r: For some reason, I was only able to override this from voto5 if it's an `async`, I don't know why. To not spend times understanding the working code, I decided to keep it async :D

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
safydy-r: Agree here, thanks for noticing. I renamed them & updated the doc too, 🙏

PR #243, opened 2022-05-17
VMO-5874 Fix console errors

object-required:
For some reason, `$router.push` doesn't return a promise in Voto5. I rewrote these two calls using async/await syntax because it handles such cases without visible problems.


bulatgab:
I remember a similar issue in the past, see https://viamoinc.atlassian.net/browse/VMO-3121
It's happening because we're using router v2 in `voto5` and v3 in `flow-builder` and they have different interfaces 😉

bulatgab (approved):

PR #253, opened 2022-05-25
VMO-6013 Update OpenResponse UI

object-required:
Removed _Maximum Response Characters_
Added _End Response Digits_

image


src/components/common/DigitsEditor.vue
```diff
@@ -0,0 +1,70 @@
+<template>
```
safydy-r, line 1: @object-required , I think `KeyDigitsSelector` name would fit better, your thoughts?

- [ ] rename if you agree

src/components/common/DigitsEditor.vue
```diff
@@ -0,0 +1,70 @@
...
+export class DigitsEditor extends mixins(Lang) {
+  @Prop({type: String, required: true}) label!: string;
+  @Prop({type: String, required: true}) value!: string;
+  @Prop({type: String, default: ''}) placeholder!: string;
+  @Prop({type: Boolean, default: true}) validState!: boolean;
+
+  get options(): MultiselectOption[] {
+    return [
+      {
+        name: KEYS,
```
safydy-r, line 42: Thanks, I didn't even know we use `1234567890#*` for `any key` in voto5 :D

src/components/common/DigitsEditor.vue
```diff
@@ -0,0 +1,70 @@
...
+    ]
+  }
+
+  get modelValue(): MultiselectOption | undefined {
+    return this.options.find((option) => this.value.includes(option.name))
+  }
+
+  set modelValue(value: MultiselectOption | undefined) {
+    if (value) {
+      this.$emit('input', value.name)
```
safydy-r, line 58: Curious, what if the value is `undefined`, shouldn't we emit the event too ?
If that's the case, then this would be `this.$emit('input', value?.name)`

src/components/interaction-designer/block-editors/EndRecordingDigitsEditor.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div class="end-recording-digits-editor">
+    <validation-message
+      :message-key="`block/${block.uuid}/config/ivr/end_recording_digits`">
+      <template #input-control="{ isValid }">
+        <div v-if="hasIvr">
```
safydy-r, line 6: Can we move the `v-if="hasIvr"` in the root element ? That way we won't render anything if `hasIvr` is false. 
I think we should have done that for `/block-editors/MaxDurationSecondsEditor.vue`, can I ask you to fix `MaxDurationSecondsEditor.vue` too ?

Context:
- before we didn't have that wrapper `<validation-message>`
- so, when we wrapped everything, we forgot to move such condition into the root element

src/components/interaction-designer/block-editors/EndRecordingDigitsEditor.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <div class="end-recording-digits-editor">
+    <validation-message
```
safydy-r, line 3: Curiosity, where you able to test a scenario which shows a validation message? eg: empty value

safydy-r (approved):
@object-required , good job here. I'm pre-approving, so you can move this to QA once you've addressed my comment. Trusting your decision on them.

src/components/interaction-designer/block-editors/EndRecordingDigitsEditor.vue
safydy-r: Pls, never mind about the MaxDurationSecondsEditor, I will fix it in my vmo-4326. Sorry

safydy-r: (Pls ignore the **deleted** comment after this in case you received it in your email)

src/components/common/DigitsEditor.vue
object-required: That does make sense, although `value` is unlikely to be _undefined_ as long as `preselect-first` is true. 
I changed the code as suggested 😉

src/components/interaction-designer/block-editors/EndRecordingDigitsEditor.vue
object-required: We don't have restrictions on `ivr.end_recording_digits` apart from _string_ type. I tried forcing a value using devtools, and the message looks as follows:

image

yeboaasare (approved):

PR #256, opened 2022-05-30
VMO-6018 Fix OpenResponse block config updates

object-required:
Now _Max Duration_ does not override _End Recording Digits_ and vice versa.


src/store/flow/block-types/MobilePrimitives_OpenResponseBlockStore.ts
```diff
@@ -24,22 +25,23 @@ const actions: ActionTree<IEmptyState, IRootState> = {
```
bulatgab, line 25: We can alternatively:
- use `block_updateConfigByPath` instead of `block_updateConfigByKey` to change individual config values directly
- remove return

bulatgab (approved):

src/store/flow/block-types/MobilePrimitives_OpenResponseBlockStore.ts
safydy-r: Good catch

PR #257, opened 2022-05-31
VMO-6083 Fix NumericBlock validation

object-required:
We tinker with `ajvErrors` to inject custom error messages. This PR adds explicit handling of such cases.

I also went ahead and rearranged checks specific for _NumericBlock_ in `MobilePrimitives_NumericResponseBlockStore.ts` so that we don't have multiple messages for a single field.


src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
```diff
@@ -72,70 +109,58 @@ const actions: ActionTree<IEmptyState, IRootState> = {
...
-            message: lang.trans('flow-builder-validation.numeric-block-min-value-must-be-lower-than-implied-max-digits'),
-          } as ErrorObject)
-          validationResults.ajvErrors.push({
-            dataPath: '/config/validation_minimum',
-            message: lang.trans('flow-builder-validation.numeric-block-min-value-must-be-lower-than-implied-max-digits'),
-          } as ErrorObject)
-        }
+      // Both validationMin and validationMax must have no more digits than allowed
+      if (hasMaxDigits) {
+        const maxValue = 10 ** maxDigits - 1
```
object-required, line 144: I noticed having `maxDigits = 2` the validation won't accept values over `18` which doesn't feel right.

cc: @pascaleli

object-required (commented):
I'm uncertain if we'd better to extract the validation logic specific for _Numeric Block_ to a separate file because it's quite complicated.

bulatgab (approved):

src/store/validation/index.ts
```diff
@@ -97,7 +97,7 @@ export const actions: ActionTree<IValidationState, IRootState> = {
       ...status,
       ajvErrors: getLocalizedAjvErrors(key, status.ajvErrors),
     })
-    if (status.ajvErrors === null) {
+    if (status.ajvErrors === null || status.ajvErrors.length === 0) {
```
bulatgab, line 100: We could alternatively return null from `errors.list()` to match the ajv interface
(and change logic for `validationResults.ajvErrors`)

src/store/flow/block-types/MobilePrimitives_NumericResponseBlockStore.ts
safydy-r: @object-required , great catch here. Thank you very much 🙏

safydy-r: @object-required , great catch here. Thank you very much 🙏

PR #258, opened 2022-06-02
VMO-6059 - feature/VMO-6059/message-block-ui-integration

bulatgab:


object-required (approved):

safydy-r:
@bulatgab , when checking this branch, I'm noticing there is a pbm on the validation:
- all the validation list is gone when we provide a block title
I think we need fix this in current ticket even we haven't noticed during ticket creation.
vmo-6059 - issue on validation

- [ ] fix validation as described above

.eslintrc.js
```diff
@@ -360,6 +360,7 @@ module.exports = {
     'no-else-return': ['off'],
     'spaced-comment': 'off',
     'line-comment-position': ['error', {position: 'above'}],
+    'arrow-parens': 'off',
```
bulatgab, line 363: This allows us to get rid of parenthesis around a single parameter in arrow functions, for example:
```
resources.some((resource) => ...)
```
becomes
```
resources.some(resource => ...)
```

src/components/interaction-designer/toolbar/ErrorNotifications.vue
```diff
@@ -111,26 +111,19 @@ export class ErrorNotifications extends mixins(Routes, Lang) {
     return pickBy(this.validationStatuses, (value: IValidationStatus, key) => key.includes('block/') && blocksMap[key])
   }
 
-  get resourceValidationStatusesForActiveFlow(): Dictionary<IValidationStatus> {
-    return pickBy(this.validationStatuses, (status, key) => {
-      const resourceUuid = replace(key, 'resource/', '')
-      return this.resourceUuidsOnActiveFlow.includes(resourceUuid) && !get(this.validationStatuses, `resource/${resourceUuid}`)?.isValid
-    })
-  }
-
```
bulatgab, line 120: I think the issue here was using `this` in an arrow function in a class's method which gave it a wrong scope (`this` was undefined)

src/components/interaction-designer/toolbar/ErrorNotifications.vue
```diff
@@ -111,26 +111,19 @@ export class ErrorNotifications extends mixins(Routes, Lang) {
...
-      if (has(this.resourceValidationStatusesForActiveFlow, `resource/${uuids[i]}`)) {
-        return true
-      }
-    }
-    return false
+    return uuids.some(uuid => {
+      const isValid = this.validationStatuses[`resource/${uuid}`]?.isValid
+      const doesBelongToActiveFlow = this.activeFlow?.resources.some(resource => resource.uuid === uuid)
+
+      return isValid === false && doesBelongToActiveFlow
```
bulatgab, line 125: `isValid` can be not only boolean, so a strict comparison with `false` is required

bulatgab:
Hi @safydy-r,
Thanks for pointing out that issue!
I've fixed it in this commit and left some comments:
https://github.com/FLOIP/flow-builder/pull/258/commits/92c9f5187f054056e076d00abf1425819c8111a0

src/components/interaction-designer/toolbar/ErrorNotifications.vue
safydy-r: Thanks for this, I like the usage of `some()`, which make an early return:
> As soon as one `true` is returned, `.some()` will itself return `true`

safydy-r (approved):
Thanks @bulatgab for fixing this 💯

PR #263, opened 2022-06-17
Feature/vmo 6021/update block exits part1

safydy-r:
[Overview]
vmo-6021 - Update block exits - part1

[For QA]
- those blocks will have 01 single `Default` exit: Message, OpenResponse, SetGroupMembership
- NumericResponse block will have 02 exits `Valid/Invalid` instead of `1/Default`, and when you switch to `Advanced mode` you will see this expression `@ISNUMBER(block.value)` associated with the `Valid` exit

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@floip/flow-builder",
-  "version": "0.4.7",
+  "version": "0.4.8",
   "description": "flow-builder provides an open source frontend for authoring flow interoperability project specification compliant flows.",
   "main": "./dist/flow-builder.common.js",
   "files": [


src/store/flow/block-types/SmartDevices_LocationResponseBlockStore.ts
```diff
@@ -34,12 +35,22 @@ const actions: ActionTree<IEmptyState, IRootState> = {
...
       accuracy_threshold_meters: 5.0,
       accuracy_timeout_seconds: 120,
+      ...await dispatch('initiateExtraVendorConfig'),
     }
+
+    props.exits = [
+      await dispatch('flow/block_createBlockDefaultExitWith', {
+        props: ({
+          uuid: await (new IdGeneratorUuidV4()).generate(),
+          name: 'Invalid',
```
safydy-r, line 45: Took the opportunity to update LocationResponse branch too, to have `Valid/Invalid` exits instead of `1/Default`

src/store/flow/block-types/SmartDevices_PhotoResponseBlockStore.ts
```diff
@@ -14,12 +15,23 @@ const actions: ActionTree<IEmptyState, IRootState> = {
...
     props.config = {
       prompt: blankMessageResource.uuid,
+      ...await dispatch('initiateExtraVendorConfig'),
     }
+
+    props.exits = [
+      await dispatch('flow/block_createBlockDefaultExitWith', {
+        props: ({
+          uuid: await (new IdGeneratorUuidV4()).generate(),
+          name: 'Invalid',
```
safydy-r, line 25: Took the opportunity to update PhotoResponse branch too, to have `Valid/Invalid` exits instead of `1/Default`

src/components/interaction-designer/block-types/MobilePrimitives_MessageBlock.vue
```diff
@@ -64,9 +64,6 @@ class MobilePrimitives_MessageBlock extends mixins(Lang) {
 
   @flowVuexNamespace.Getter resourcesByUuidOnActiveFlow!: { [key: string]: IResource }
 
-  @flowVuexNamespace.Action block_convertExitFormationToUnified!:
```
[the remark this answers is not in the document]
bulatgab: good catch 👍

src/store/flow/block-types/SmartDevices_LocationResponseBlockStore.ts
```diff
@@ -34,12 +35,22 @@ const actions: ActionTree<IEmptyState, IRootState> = {
       prompt: blankMessageResource.uuid,
       accuracy_threshold_meters: 5.0,
       accuracy_timeout_seconds: 120,
+      ...await dispatch('initiateExtraVendorConfig'),
     }
+
+    props.exits = [
+      await dispatch('flow/block_createBlockDefaultExitWith', {
+        props: ({
+          uuid: await (new IdGeneratorUuidV4()).generate(),
```
object-required, line 44: Not a review comment, but rather a general thought. Would it be better to reuse an instance of generator from `flow-runner` context?

By importing `IdGeneratorUuidV4` explicitly, we can possibly override user's preference on ID generation:

https://github.com/FLOIP/flow-runner/blob/master/src/flow-spec/IContext.ts#L137

Besides, as the `generate()` is not static, the generator is semantically allowed to have some state, so creating new instances of it may lead to duplicates:

```ts
class IdGeneratorSimple implements IIdGenerator {
  let id = 0
  async generate(): Promise<string> {
    return ++this.id
  }
}
```

object-required (approved):

src/store/flow/block-types/SmartDevices_LocationResponseBlockStore.ts
safydy-r: Thanks @object-required for you input here, sorry I don't really understand the proposal :D .
Currently, we're using `IdGeneratorUuidV4` https://github.com/FLOIP/flow-runner/blob/e7a3ecf5d27bc941bff70dc69a5ffa578fc227cf/src/domain/IdGeneratorUuidV4.ts#L28 
```
export class IdGeneratorUuidV4 implements IIdGenerator {
  async generate(): Promise<string> {
    return uuid.v4()
  }
}
```
which provides a very unique uuid for every generation with `uuid.v4()`, so no duplicates expected !

(moving this PR to QA, but let's have a call about this if needed ;) )

PR #265, opened 2022-06-27
VMO-6346 Block code (name) editor behavior

object-required:
The `block.name` is updated when user edits either _block title_ or _block code_.

Initially, I had the idea to leave `block.name` empty until a user explicitly changes it, but it turned impractical as both `IBlock` and JSON Schema define `name` attribute mandatory.

This PR adds a new field to `block.ui_metadata.has_user_defined_name` to mark block names edited by user. The newly created `block_setName` action contains related conditions so that we don't overwrite `name` together with `label`.

See demo 🎞️ in [VMO-6346](https://viamoinc.atlassian.net/browse/VMO-6346)


bulatgab (approved):

src/store/flow/block.ts
```diff
@@ -122,9 +130,32 @@ export const mutations: MutationTree<IFlowsState> = {
...
+  },
+
+  block_setName(
+    {commit, state},
+    {blockId, value, isUserDefinedName = false}: {blockId: IBlock['uuid'], value: string, isUserDefinedName: boolean},
+  ) {
+    const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
+    const canSetName = block.ui_metadata?.has_user_defined_name !== true
+
+    if (canSetName || isUserDefinedName) {
```
safydy-r, line 145: @object-required , If I'm not wrong, the logic here is `if the user hasn't defined it yet, then auto update the name from the payload value`, right ?

With that, we know that the **final state** we want to know is "should we auto update the name?". What if we use that final state to rename the flag by using:
- `should_auto_update_name` instead of `has_user_defined_name`
- `block_autoUpdateName`  instead of `block_setUserDefinedName`
- `shouldAutoUpdateName` instead of `canSetName` (which is more like permission name). Or we can even avoid creating a variable for that, it's only used once.

That would be easier to read I think, your thoughts?

- [ ] rename the flag if you agree

src/store/flow/block.ts
```diff
@@ -17,6 +17,14 @@ export const mutations: MutationTree<IFlowsState> = {
     findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
       .name = value
   },
+  block_setUserDefinedName(state, {blockId, value}) {
+    const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
+
+    block.ui_metadata = {
```
safydy-r, line 23: Let's use the same approach to update config & vendor_metadata `by path`, eg: [block_updateConfigByPath](https://github.com/FLOIP/flow-builder/blob/master/src/store/flow/block.ts#L94).

- [ ] Create a new mutation `block_updateUiMetadataByPath` in https://github.com/FLOIP/flow-builder/blob/master/src/store/flow/block.ts, and call it here.

src/components/interaction-designer/block-editors/NameEditor.vue
```diff
@@ -63,16 +63,38 @@ export class NameEditor extends mixins(Lang) {
...
     if (e.key.match(/\W+|Enter/g)) {
       e.preventDefault()
     }
   }
 
-  @flowVuexNamespace.Mutation block_setName!: ({blockId, value}: {blockId: IBlock['uuid'], value: IBlock['name']}) => void
+  handleCompleteEditing(): void {
+    this.editBlockName = false
+
+    if (this.blockName === '') {
```
safydy-r, line 81: good catch

src/components/interaction-designer/block-editors/NameEditor.vue
```diff
@@ -63,16 +63,38 @@ export class NameEditor extends mixins(Lang) {
...
+  }
+
+  @flowVuexNamespace.Action block_setName!: ({
+    blockId,
+    value,
+    isUserDefinedName,
+  }: {
+    blockId: IBlock['uuid'],
+    value: IBlock['name'],
+    isUserDefinedName: IBlockUIMetadata['has_user_defined_name'],
```
safydy-r, line 94: @object-required , can we get the type of `IBlockUIMetadata['has_user_defined_name']` if we didn't update flow-runner interface `IBlockUIMetadata` to have `has_user_defined_name` ?
Or did you add it but forgot to share the PR ?

safydy-r: - [ ] update `IBlockUIMetadata` with the new flag (optional boolean)

safydy-r (changes requested):
In general I'm good, left several questions/requests though.

src/components/interaction-designer/block-editors/NameEditor.vue
object-required: Added to https://github.com/FLOIP/flow-runner/pull/78

package.json
```diff
@@ -28,7 +28,7 @@
     "@avcs/autosuggest": "^1.7.5",
     "@floip/expression-evaluator": "^2.1.0",
     "@floip/expression-parser": "^1.11.3",
-    "@floip/flow-runner": "1.0.0-rc4.6",
+    "@floip/flow-runner": "https://github.com/FLOIP/flow-runner.git#feature/VMO-6346/block-code",
```
safydy-r, line 31: - [ ] use the published version from npm after flow-runner approval

src/store/flow/block.ts
```diff
@@ -137,9 +158,40 @@ export const mutations: MutationTree<IFlowsState> = {
...
+    dispatch('block_setName', {blockId, value: snakeCase(value)})
+  },
+
+  block_setName(
+    {commit, state},
+    {blockId, value, lockAutoUpdate = false}: {blockId: IBlock['uuid'], value: string, lockAutoUpdate: boolean},
+  ) {
+    const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
+
+    // lockAutoUpdate = true means user overrides auto-generated name
```
safydy-r, line 172: - [ ] move this  in JS Doc, above block_setName definition

src/store/flow/block.ts
```diff
@@ -108,6 +108,27 @@ export const mutations: MutationTree<IFlowsState> = {
 
     Vue.set(pointer, chunks[0], value)
   },
+  block_updateUIMetadataByPath(state, {blockId, path, value}: {blockId: string, path: string, value?: object | string | number | boolean}) {
```
safydy-r, line 111: 💯

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -472,6 +472,7 @@ export class TreeBuilderToolbar extends mixins(Routes, Permissions, Lang) {
       type,
       ui_metadata: {
         canvas_coordinates: computeBlockUiData(this.activeBlock),
+        should_auto_update_name: true,
```
safydy-r, line 475: - [ ] make this optional as in flow-runner recommendation https://github.com/FLOIP/flow-runner/pull/78/files#diff-8e0400228dd698ac38c3bdcccec67c81d0c45614354d8cf60f4e6af3dd863aebR52

safydy-r (approved):
@object-required , I'm pre-approving with one concern:
- the `should_auto_update_name` should optional I think, but let me know if you disagree

object-required:
> @object-required , I'm pre-approving with one concern:
> 
> * the `should_auto_update_name` should optional I think, but let me know if you disagree

I see you addressed this concern in https://github.com/FLOIP/flow-runner/pull/78/commits/ba6152209f77e8da7f7b1d5fb763ad34e619bad2, thank you for taking action here 🙏

PR #268, opened 2022-07-01
VMO-6311 Multiple choice

object-required:
This is a complementary PR for https://bitbucket.org/voto/voto5/pull-requests/2527/vmo-6311-multiple-choice


src/store/flow/block-types/Core_SetContactPropertyStore.ts
```diff
@@ -4,11 +4,17 @@ import {IBlockConfig} from '@floip/flow-runner'
 import {cloneDeep} from 'lodash'
 import BaseStore, {actions as baseActions, IEmptyState} from '@/store/flow/block-types/BaseBlock'
 
+export interface IMultipleChoiceOption {
```
safydy-r, line 7: I'd suggest we rename this with something like `IContactPropertyMultipleChoice` because the word `MultipleChoice` could be confusing as we have the internal vocabulary `MultipleChoiceQuestion`.
(with the prefix `I`, we can remove the last postfix `Option` I think)

Your thoughts?

- [ ] rename or reach out to me if you have a concern

safydy-r (approved):
Pre-approving this PR with one small request

PR #273, opened 2022-07-09
VMO-6445 SetGroupMembership

object-required:


src/lib/validations/Core_SetGroupMembershipValidator.ts
```diff
@@ -0,0 +1,15 @@
+import {IBlock, IBlockConfig} from '@floip/flow-runner'
+import {ExtendedValidatorBase, ValidationResults} from './ExtendedValidatorBase'
+
+export default class Core_SetGroupMembershipValidator extends ExtendedValidatorBase {
+  static runProgrammaticValidations(block: IBlock<IBlockConfig>, options?: unknown): ValidationResults | null {
+    const {groups, is_member} = block.config
+
+    if (is_member !== undefined && groups.length === 0) {
```
safydy-r, line 8: - [ ] use `groups?.length` as groups could be undefined

src/lib/validations/ExtendedValidatorBase.ts
```diff
@@ -0,0 +1,44 @@
...
+export type ValidationMessageSuffix = string
+export type ValidationResult = [ErrorObject['dataPath'], ValidationMessageSuffix]
+export type ValidationResults = ValidationResult[]
+
+/**
+ * This class encapsulates JSON schema validation logic
+ * and also runs user-defined programmatic checks from within subclasses.
+ */
+export class ExtendedValidatorBase {
+  static runProgrammaticValidations(_block: IBlock, _options?: unknown): ValidationResults | null {
```
safydy-r, line 17: Thanks for using this naming

src/lib/validations/Core_SetGroupMembershipValidator.ts
safydy-r: Done

safydy-r (approved):
Instead of requesting changes here, I made it myself. Pls check my last commit to check what I've changed. Not big changes though.
Thanks @object-required

PR #275, opened 2022-07-12
Feature/vmo 6421/vmo 6454/update generic scp for all block config

safydy-r:
**Note:**
- I branched from 6428, because there were many ui refactoring there.

Overview:
vmo-6454-update-generic-scp-ui


src/store/flow/block.ts
```diff
@@ -162,6 +170,80 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
     })
   },
 
+  /**
+   * Make sure we have Array set_contact_property,
+   * if it's not an array, then the mutation will set an object like {set_contact_property: {0: object1, 1: object2}} instead
+   */
+  block_makeSureSetContactPropertyExistsAsArray({commit, state}, {blockId}: { blockId: string}) {
```
safydy-r, line 177: @object-required , I believe these are the changes you wanted in the 3rd subtasks https://viamoinc.atlassian.net/browse/VMO-6461 ?

object-required: Initially, I was thinking of using `property_key` to address an element of the array, but now I realize we definitely need to use numeric indexes for the community version.

I will add the `property_key` based actions in [VMO-6461](https://viamoinc.atlassian.net/browse/VMO-6461), and we will need both implementations in [VMO-6233](https://viamoinc.atlassian.net/browse/VMO-6233).

PR #276, opened 2022-07-12
VMO-6456 update path & set_contact_property references in json, stori…

safydy-r:
Overview
Screen Shot 2022-07-12 at 12 54 05 PM


src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
```diff
@@ -36,7 +36,7 @@
 
       <validation-message
         #input-control="{ isValid }"
-        :message-key="`block/${block.uuid}/config/set_contact_property/property_key`">
+        :message-key="`block/${block.uuid}/config/set_contact_property/0/property_key`">
```
object-required, line 39: I think `0` will be replaced with`x` in

https://github.com/FLOIP/flow-builder/blob/feature/VMO-6421/VMO-6456/update-core-scp-block/src/store/validation/validationHelpers.ts#L94-L97

object-required (approved):

src/components/interaction-designer/block-editors/ContactPropertyEditor.vue
safydy-r: Thanks for this @object-required , I will change

PR #279, opened 2022-07-13
 VMO-6481 Fix Case Block Validation

object-required:
This PR
- Adds default `config` value in `BaseBlock` to avoid validation errors.
- Sets `should_auto_update_name` in `Core_CaseBlockStore`

ℹ️ The target branch is `feature/VMO-6428/update-ui-based-on-vendor_metadata-ui_metadata-schema-change`


safydy-r (commented):
@object-required , thanks for providing this, this reminds me a [TODO](https://github.com/FLOIP/flow-builder/blob/master/src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue#L482) we could address. Give me a moment, I'm creating a new improved PR I'd like to suggest to fix this.

PR #280, opened 2022-07-13
VMO-6461 SetContactProperty Array Methods

object-required:
This PR targets https://github.com/FLOIP/flow-builder/pull/275 because it adds functionality and is not an independent development.


src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
...
+    const block = findBlock(blockId, state as unknown as IContext)
+
+    // Ensure the block has a set_contact_property[]
+    if (!block.config.set_contact_property) {
+      block.config.set_contact_property = []
+    }
+
+    // We should avoid duplicates
+    if (block.config.set_contact_property.find(({property_key}) => property_key === key)) {
+      throw new Error(`set_contact_property[${key}] already exists`)
```
bulatgab, line 26: Would it be benefitial to add a one-line doc to `block_addContactProperty` about this?
So that wherever we consume it, we know that it might throw an error.

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface ISetContactPropertyIndexWithValue extends ISetContactPropertyIndex {
+  value: SetContactProperty['property_value'],
+}
+
+export const mutations: MutationTree<IFlowsState> = {
+  block_addContactProperty(state, {blockId, key, value}: ISetContactPropertyIndexWithValue) {
+    const block = findBlock(blockId, state as unknown as IContext)
+
+    // Ensure the block has a set_contact_property[]
```
bulatgab, line 19: The code is self-explanatory (to me). I vote for removing these comments.
@safydy-r do you agree? Or should we keep them?

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (block.config.set_contact_property.find(({property_key}) => property_key === key)) {
+      throw new Error(`set_contact_property[${key}] already exists`)
+    }
+
+    block.config.set_contact_property = [
+      {
+        property_key: key,
+        property_value: value,
+      },
+      ...block.config.set_contact_property ?? [],
```
bulatgab, line 34: `set_contact_property` is already ensured to be defined. Would you like to remove one of them?

bulatgab: Quick question: we're not `push`ing to the array because we want Vue reactivity, right? (and pushing would merely mutate the existing array)

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
...
+  },
+}
+
+export const actions: ActionTree<IFlowsState, IRootState> = {
+  block_getContactPropertyByKey(state, {blockId, key}: ISetContactPropertyIndex) {
+    return findBlock(blockId, state as unknown as IContext)
+      .config
+      .set_contact_property
+      ?.find(({property_key}) => property_key === key)
+      ?.property_value ?? undefined
```
bulatgab, line 62: `?? undefined` is probably not needed

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
...
+      commit('block_addContactProperty', {blockId, key, value})
+    } else {
+      commit('block_changeContactProperty', {blockId, key, value})
+    }
+  },
+
+  block_deleteContactPropertyByKey({commit}, {blockId, key}: ISetContactPropertyIndex) {
+    commit('block_removeContactProperty', {blockId, key})
+  },
+}
```
bulatgab, line 76: Could you also add return types, please?

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
...
+import {IFlowsState} from '../index'
+
+export interface ISetContactPropertyIndex {
+  blockId: IBlock['uuid'],
+  key: SetContactProperty['property_key'],
+}
+
+export interface ISetContactPropertyIndexWithValue extends ISetContactPropertyIndex {
+  value: SetContactProperty['property_value'],
+}
```
bulatgab, line 13: Nice work with types and the whole job with the new add/remove/set/change interface 👍

src/store/flow/block/set-contact-property.ts
bulatgab: Bonus question: would it be more 'intuitive' to add the new element at the end of the array?

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
...
+      .filter(({property_key}) => property_key !== key)
+  },
+
+  block_changeContactProperty(state, {blockId, key, value: newValue}: ISetContactPropertyIndexWithValue) {
+    const block = findBlock(blockId, state as unknown as IContext)
+
+    block.config.set_contact_property = (block.config.set_contact_property ?? [])
+      .map(({property_key, property_value: oldValue}) => ({
+        property_key,
+        property_value: property_key === key ? newValue : oldValue,
```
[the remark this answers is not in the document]
bulatgab: Minor suggestion: should we console log if the element is not found and the function call has no effect?
I don't think it's going to be used anywhere else other than `block_setContactPropertyByKey`, so this maybe overkill. Leaving the decision up to you 😉

src/store/flow/block/set-contact-property.ts
safydy-r: I vote to remove it as well, thanks

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,76 @@
+import {IRootState} from '@/store'
```
safydy-r, line 1: @object-required , thnx for creating this file. I'd suggest we move here all setContactProperty actions I created in `block.ts`, what do you think ?

https://github.com/FLOIP/flow-builder/blob/feature/VMO-6461/scp-array-methods/src/store/flow/block.ts#L178-L251

safydy-r (changes requested):
Good changes here. I have one single major concern, the confusion on `index` usage. Let me know if you have any questions.

src/store/flow/block/set-contact-property.ts
object-required: Actually, I meant to suggest doing the other way around. The `block.ts` almost hit 400 lines, and splitting its contents by grouping methods logically could make it easier to maintain, just like we did with routes recently.

src/store/flow/block/set-contact-property.ts
object-required: Although it _should_ be defined when we invoke this action, it may not be in case of errors because it's not required according to typings:

```ts
export interface IBlockConfig extends Record<string, any> {
  set_contact_property?: SetContactProperty[]
}
```

Unless you insist, I would be more comfortable keeping the extra fallback instead of `block.config!.set_contact_property` 😉 

We have to prepend newly added properties to make them appear at all, this will have to change if we decide to show/edit all properties; [VMO-6509](https://viamoinc.atlassian.net/browse/VMO-6509).

src/store/flow/block/set-contact-property.ts
```diff
@@ -0,0 +1,150 @@
...
+import {findBlockOnActiveFlowWith as findBlock, IBlock, IContext, SetContactProperty} from '@floip/flow-runner'
+import {ActionTree, MutationTree} from 'vuex'
+import {IFlowsState} from '../index'
+
+export interface ISetContactPropertyBlockKey {
```
safydy-r, line 6: Thanks for renaming ;)

safydy-r (approved):
@object-required , this LGTM.

PR #281, opened 2022-07-13
Feature/vmo 6481/fix case block validation  with more refactoring

safydy-r:
Note:
- inspired from @object-required 's PR https://github.com/FLOIP/flow-builder/pull/279
- by trying to set default props at the block creation as possible

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@floip/flow-builder",
-  "version": "0.4.11",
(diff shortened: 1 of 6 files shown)


safydy-r (commented):
remaining todo @object-required , 

- [ ] build it before merge so we can consumer from voto5

object-required (approved):

PR #284, opened 2022-07-14
VMO-6233 Integration Branch

object-required:


src/store/flow/block-types/Core_CaseBlockStore.ts
```diff
@@ -16,6 +16,7 @@ const actions: ActionTree<IEmptyState, IRootState> = {
       floip: {
         ui_metadata: {
           branching_type: OutputBranchingType.ADVANCED,
+          should_auto_update_name: true,
```
safydy-r, line 19: I think we don't need this anymore, as every block will have it from baseBlock https://github.com/FLOIP/flow-builder/blob/master/src/store/flow/block-types/BaseBlock.ts#L42

- [ ] double check if we can remove this, then remove it

bulatgab (approved):

safydy-r (approved):

PR #287, opened 2022-07-15
Feature/vmo 6309/fix get property action

bulatgab:
Fix issues that I had in voto5:
- turn `block_getContactPropertyOnIndex` and `block_getContactPropertyValueByKey` into getters which are synchronous. (They were actions and were returning promises. We couldn't await there)
- use double arrow function because we need to provide params (with regular getters we don't use params). This is called [method-style access](https://vuex.vuejs.org/guide/getters.html#method-style-access)

### DIFF
--- a/src/store/flow/block.ts
+++ b/src/store/flow/block.ts
@@ -8,7 +8,7 @@ import {
 } from '@floip/flow-runner'
 import {ActionTree, GetterTree, MutationTree} from 'vuex'
 import {IRootState} from '@/store'
-import {defaults, get, has, isArray, isNil, last, reject, snakeCase, toPath} from 'lodash'
+import {defaults, has, isArray, isNil, last, reject, snakeCase} from 'lodash'
 import {IdGeneratorUuidV4} from '@floip/flow-runner/dist/domain/IdGeneratorUuidV4'
 import {IFlowsState} from '.'
 import {removeBlockValueByPath, updateBlockValueByPath} from './utils/vuexBlockHelpers'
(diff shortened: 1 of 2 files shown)


src/store/flow/block.ts
```diff
@@ -266,7 +268,10 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
    * - Valid
    * - Invalid
    */
-  async block_updateBranchingExitsWithInvalidScenario({state, dispatch}, {blockId, test}: {blockId: IBlock['uuid'], test: IBlockExit['test']}) {
+  async block_updateBranchingExitsWithInvalidScenario(
+    {state, dispatch},
+    {blockId, test}: {blockId: IBlock['uuid'], test: IBlockExit['test']},
+  ) {
```
bulatgab, line 274: eslint fix

src/store/flow/block.ts
```diff
@@ -8,14 +8,16 @@ import {
 } from '@floip/flow-runner'
 import {ActionTree, GetterTree, MutationTree} from 'vuex'
 import {IRootState} from '@/store'
-import {defaults, get, has, isArray, isNil, last, reject, snakeCase, toPath} from 'lodash'
+import {defaults, has, isArray, isNil, last, reject, snakeCase} from 'lodash'
```
bulatgab, line 11: eslint fix

src/store/flow/block/set-contact-property.ts
```diff
@@ -54,90 +72,71 @@ export const mutations: MutationTree<IFlowsState> = {
 }
 
 export const actions: ActionTree<IFlowsState, IRootState> = {
-    /**
+  /**
    * Make sure we have Array set_contact_property,
    * if it's not an array, then the mutation will set an object like {set_contact_property: {0: object1, 1: object2}} instead
    */
```
bulatgab, line 78: 👇 fix indentation

object-required (approved):
I didn't know component getters cannot be async 🤔 
LGTM 👍

PR #290, opened 2022-07-20
Feature/vmo 6076/fix z index of block config

safydy-r:


src/components/interaction-designer/Block.vue
```diff
@@ -10,7 +10,7 @@
     <plain-draggable
       v-if="hasLayout"
       ref="draggable"
-      class="block"
+      class="block-draggable"
```
safydy-r, line 13: This is the main reason we had that bug, because we're using `block` class in the root element too https://github.com/FLOIP/flow-builder/pull/290/files#diff-43f7c5f1a1252fa694baadbe4e2a0ec2f48c3a9041968f25ed1cfab7b4bdff9aR3

object-required (approved):
Nicely done ✨

bulatgab:
It's nice to see things fixed 👍

bulatgab (approved):

PR #293, opened 2022-07-25
Feature/vmo 6231/viamo UI for run flow block

bulatgab:
Make empty flow_id invalid in run flow block:

Screen Shot 2022-07-26 at 1 42 20 AM


object-required (approved):

src/lib/validations/Core_RunFlowBlockValidator.ts
```diff
@@ -0,0 +1,16 @@
+import {IBlock} from '@floip/flow-runner'
+import {ExtendedValidatorBase, ValidationResults} from './ExtendedValidatorBase'
+
+export default class Core_RunFlowBlockValidator extends ExtendedValidatorBase {
+  static runProgrammaticValidations(block: IBlock, options?: unknown): ValidationResults | null {
+    const {flow_id} = block.config
+
+    if (typeof flow_id !== 'string' || flow_id.length === 0) {
```
safydy-r, line 8: as flow_id is a REQUIRED string https://github.com/FLOIP/flow-runner/blob/master/src/model/block/IRunFlowBlockConfig.ts#L23, why don't we provide `undefined` as value when it's empty instead of `empty string`. That way we don't need a programmatic validation.
Your thoughts?

safydy-r (commented):
Thanks for this @bulatgab , I only have one main concerns related to the usage of programmatic validation

src/store/flow/block.ts
```diff
@@ -110,9 +104,8 @@ export const mutations: MutationTree<IFlowsState> = {
     if (!block.ui_metadata) {
       Vue.set(block, 'ui_metadata', {})
     }
-    const base = block.ui_metadata!
 
-    let pointer = base
+    let pointer = block.ui_metadata!
```
bulatgab, line 108: `base` wasn't used, so I simplified (and eslint was suggesting it as well, as I remember it)

src/lib/validations/Core_RunFlowBlockValidator.ts
bulatgab: Maybe, yes, thanks Safydy!

Could you maybe help a bit?
If I'm doing this:
```
async createWith(...) {
    ...
    props.config = {
      flow_id: null,
    }
    ...
  }
```
I'm getting this:
Screen Shot 2022-07-26 at 4 37 37 PM

bulatgab: If I'm doing this
```
props.config = {
      flow_id: undefined,
}
```
or just `props.config = {}`
then I'm getting this
Screen Shot 2022-07-26 at 4 41 58 PM

bulatgab: The second one is worse, I think. Because its path is `/config` which is not too specific and it's not showing the red box.

And in the first case it's `/config/flow_id` which is specific and which shows the red box.

bulatgab: I wonder how we can customise the validation message...

bulatgab: Trying all this here:
https://github.com/FLOIP/flow-builder/pull/295/files

safydy-r:
Declining this in favour of the new PR: https://github.com/FLOIP/flow-builder/pull/295

PR #315, opened 2022-08-11
Feature/vmo 5441/vmo 4470/remove non rc4 blocks

bulatgab:


src/assets/messages.json
```diff
@@ -90,8 +88,6 @@
     "CollaborativeFilteringRatingBlock": "Collaborative Filtering Rating",
     "CollaborativeFilteringRatioBranchBlock": "Branch via Collaborative Filtering Ratio",
     "ConnectToOperatorBlock": "Connect to Operator",
-    "ConsoleIO.Print": "Print",
-    "ConsoleIO.Read": "Read",
```
object-required, line 94: I wonder if we should also remove the strings from [the localization sheet](https://docs.google.com/spreadsheets/d/136-y0DYRWFy6Max-Xz2AsqJ95TQvwrKjal6Wn-z0lsI/edit#gid=561815572).

object-required (approved):

src/assets/messages.json
bulatgab: Good point.
@safydy-r what's your opinion?

safydy-r: @bulatgab , if we're sure those localization are not used, I'd encourage us to remove them in the localization sheet. Maintenance like this would keep the sheet small ;) 

- [ ] let's remove them

tests/unit/__snapshots__/storybook.spec.ts.snap
```diff
@@ -594,28 +594,6 @@ exports[`Storyshots InteractionDesigner/Toolbar Saving State 1`] = `
                           >
                             <!---->
                              
-                            <a
```
safydy-r, line 597: Thanks @bulatgab

safydy-r (approved):
Pre-approving with these requests:

- [x] bump the version in `package.json`, and create a git tag with that version. This will help us getting the source code in future if we need similar blocks.
- [ ] build & npm publish this after QA is DONE

bulatgab:
@safydy-r
Created [this](https://github.com/FLOIP/flow-builder/releases/tag/0.4.16) tag, let me know if it doesn't look right

src/assets/messages.json
bulatgab: I've removed those rows
(I've left a copy of them here just in case: https://docs.google.com/spreadsheets/d/1_sul1YRaMpaADcBbs2P6hk0f3LMXVVEqBJHF3fB0KWk/edit?usp=sharing)

bulatgab: I've run the script in voto5 to update the localizations but it didn't remove those lines.
I suspect that the script only adds and doesn't remove. If you're okay with it, I'd suggest we ignore this.
Maybe the script gets improved in the future.

PR #317, opened 2022-08-15
VMO-4178 Add a slot for the `Create Flow` button

object-required:
This allows overriding the default behaviour of Flow Builder less invasively.

### DIFF
--- a/src/views/ImportFlow.vue
+++ b/src/views/ImportFlow.vue
@@ -83,7 +83,7 @@
 
               <slot
                 name="createAction"
-                :container="flowContainer">
+                :json="flowJson">
                 <a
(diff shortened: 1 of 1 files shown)


safydy-r (changes requested):
Thanks for trying to find a solution here @object-required . I feel like I need to put this on hold, the time I implement https://viamoinc.atlassian.net/browse/VMO-6756. In fact, I might have similar approach which could directly work with the frontend https://viamoinc.atlassian.net/browse/VMO-6755 .

- [x] wait for Safidy

safydy-r:
as discussed with SBN, we will choose this approach instead https://github.com/FLOIP/flow-builder/pull/322, so closing the PR.

PR #318, opened 2022-08-15
Feature/vmo 3952/vmo 6755/improve frontent import validation msg

safydy-r:
[Note for devs]
- the previous validation uses the flow-runner validator, which give us only one result even the flow container has multiple invalid fields
- instead of trying to fix the flow-runner validator, I switched the validator to the current `flow-builder validator module`, which provides us much more details

[overview]
vmo-6755 - frontend part


src/store/validation/index.ts
```diff
@@ -86,6 +88,18 @@ export const mutations: MutationTree<IValidationState> = {
...
+  resetValidationStatuses(state, {key}): void {
+    Vue.set(state.validationStatuses, key, {ajvErrors: undefined})
+    Vue.set(state.validationStatuses[key], 'ajvErrors', [])
+  },
+  pushAjvErrorToValidationStatuses(state, {key, ajvError}): void {
+    if (state.validationStatuses[key]?.ajvErrors === undefined) {
+      Vue.set(state.validationStatuses, key, {ajvErrors: undefined})
+      Vue.set(state.validationStatuses[key], 'ajvErrors', [])
+    }
+    // state.validationStatuses[key]!.isValid = false
```
object-required, line 100: Is this a temporary piece?

src/components/interaction-designer/flow-editors/import/TechErrorNotifications.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div class="tech-error-notification">
+    <div v-for="(ajvError, key) in validationErrors" :key="key">
+      <h6>
+        <b>#{{ key + 1 }} &nbsp;</b>
+        <b>Location</b>: {{ ajvError.dataPath }}, <b>Message</b>: {{ ajvError.message }}
```
object-required, line 6: Should we consider translating these strings?

object-required (approved):

src/store/validation/index.ts
safydy-r: removed !

src/components/interaction-designer/flow-editors/import/TechErrorNotifications.vue
safydy-r: Good catch

PR #326, opened 2022-08-18
VMO-6574 Block Connections

object-required:
VMO-6574-2


safydy-r (approved):
I have no much comment, I'm happy you found a way.
(I think we should split this giant component `Block.vue` in the future, it's becoming bigger and bigger)

safydy-r:
- [ ] Fix the cursor type (use the hand icon instead of arrow) when we're hovering the exit

safydy-r:
- [ ] fix this console when trying to drag an exit
Screen Shot 2022-08-18 at 9 22 34 AM

bulatgab (approved):
(This is beyond my current skills) Great work 👍

PR #327, opened 2022-08-18
Feature/vmo 6599/update block name in expressions

bulatgab:


src/assets/messages.json
```diff
@@ -3350,7 +3350,7 @@
     "set-group-membership-groups-required": "At least one group should be specified",
     "date-must-be-selected": "A specific date must be selected",
     "input-must-be-a-number": "Input must be a number",
-    "numeric-block-must-be-selected": "A block must be selected",
+    "numeric-block-must-be-selected": "An (existing) block must be selected",
```
bulatgab, line 3353: With this updated wording I've tried to cover both:
- that something should be selected
- and that it should be a block existing on the flow

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -42,27 +45,23 @@ export class BuilderCanvas extends Vue {
...
   async onActiveFlowChanged(newFlow: IFlow): Promise<void> {
-    console.debug('watch/activeFlow:', 'active flow has changed from builder canvas, validating ...')
-    await this.validate_flow({flow: newFlow})
+    this.debounceFlowValidation({newFlow})
   }
 
+  debounceFlowValidation = debounce(function (this: any, {newFlow}: {newFlow: IFlow}) {
+    console.debug('watch/activeFlow:', 'active flow has changed from builder canvas, validating ...')
+    this.validate_flow({flow: newFlow})
+  }, DEBOUNCE_VALIDATION_TIMER)
```
bulatgab, line 54: Debouncers are not required for the solution.
But they:
- provide a cleaner console
- give maybe a better ux (we don't validate until the user finished making the change)
- maybe improve performance

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -42,27 +45,23 @@ export class BuilderCanvas extends Vue {
...
-      newBlocks.map(async (currentNewBlock) => {
-        const currentOldBlock = find(oldBlocks, {uuid: currentNewBlock.uuid})
-        if (!isEqual(currentNewBlock, currentOldBlock)) {
-          await this.validate_block({block: currentNewBlock})
-        }
-      }),
-    )
-  }
+    this.validate_allBlocksWithinFlow()
+  }, DEBOUNCE_VALIDATION_TIMER)
```
bulatgab, line 64: For blocks I've changed it so that we are validating not only the blocks that have changed, but all blocks.
We need it because if the changed (e.g. renamed) block is referenced in a different block (e.g. in date expression) we want to invalidate that.

src/store/flow/block.ts
```diff
@@ -154,18 +154,24 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
 
   /**
    * Set block name
+   * @param commit
+   * @param dispatch
+   * @param state
```
bulatgab, line 159: eslint fix

safydy-r (commented):
Thanks @bulatgab for this PR. My day was a bit booked, I was hoping to review this but I couldn't really. I'd rely on @object-required 's review for this one. Thank you very much.

src/assets/messages.json
object-required: Alter to `An existing block must be selected` as per discussion.

src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -16,7 +16,7 @@
 
 <script lang="ts">
 import {Component, Prop, Vue, Watch} from 'vue-property-decorator'
-import {cloneDeep, debounce, find, isEqual, maxBy} from 'lodash'
+import {cloneDeep, debounce, find, isEqual, maxBy, throttle} from 'lodash'
```
object-required, line 19: Unused `throttle`

src/components/interaction-designer/BuilderCanvas.vue
object-required: I second performance considerations 👍

object-required (approved):
Pre-approving with a couple of minor fixes in mind

PR #329, opened 2022-08-19
VMO-6572 Adjust editor position

object-required:
VMO-6572-1

### DIFF
--- a/src/components/interaction-designer/Block.vue
+++ b/src/components/interaction-designer/Block.vue
@@ -442,6 +442,7 @@ export class Block extends mixins(Lang) {
     const {block} = this
     this.$nextTick(() => {
       this.setBlockPositionTo({position: {x, y}, block})
+      this.updateTranslatedBlockEditorPosition()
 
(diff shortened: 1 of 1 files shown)


safydy-r (changes requested):
Thanks @object-required for trying to find a solution here. I think, this is a tough one.

I've double checked the implementation and
- The height is working
- But I see a major UX issue because the config UI doesn't follow the block when we move it to the left / right. See this capture
vmo-6572 - PR review - round 1

My proposal would be:
- If you think this would still need more than 1day implementation, pls talk to Kyle / Nancy about what should we do: Shall we keep working on it, or shall we decrease the priority and focus on something more important ?

- [ ] fix the config position when we drag the block

safydy-r:
@object-required , I'd like to re-encourage us to always post a screen capture for complex UI / UX changes like this to avoid the reviewer to re-serve in local machine for double checking. ;) 

cc: @bulatgab

bulatgab (approved):

PR #333, opened 2022-08-26
VMO-6850 Content validation

object-required:
This change adds missing resource variants after receiving the flow back from backend.


object-required:
Combining this ticket with VMO-6855 into a single task → VMO-6875

PR #335, opened 2022-08-30
VMO-6875 MCQ Fixes

object-required:


src/store/flow/resource.ts
```diff
@@ -309,3 +310,40 @@ export function discoverContentTypesFor(mode: SupportedMode, resource?: IResourc
...
+    const lookupValues = castArray(lookup)
+
+    for (let i = 0; i < lookupValues.length; i += 1) {
+      if (allowedValues.includes(lookupValues[i])) {
+        return true
+      }
+    }
+    return false
+  }
+}
```
bulatgab, line 325: This is a smart function 🙂

But I'm thinking, is it maybe too complex for the task? (It is complex because it 1) returns another function and 2) can handle both arrays and individual elements). Maybe it's not too hard per se, but it is a new pattern we haven't used. Also, I wonder if this function will ever be used again (is it worth it).

Do you think inlining the logic and using `flow.supported_modes.some(mode => value.modes.includes(mode))` or `_.intersection([],[]).length` as alternatives could be simpler?

src/store/flow/resource.ts
```diff
@@ -309,3 +310,40 @@ export function discoverContentTypesFor(mode: SupportedMode, resource?: IResourc
...
+        values: resource.values.filter(value => (
+          hasAllowedMode(value.modes)
+          && hasSupportedLanguage(value.language_id)
+        )),
+      }))
+      .filter(resource => !isEmpty(resource.values))
+  }
+
+  return result
+}
```
bulatgab, line 349: optional: use `forEach`

bulatgab (changes requested):

src/store/flow/flow.ts
```diff
@@ -361,38 +361,34 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
-                  content_type,
-                  language_id: language,
-                  modes: [
-                    mode,
-                  ],
-                  value: '',
-                },
+    resources
+      // Choices are a special case, we should not add variants
+      .filter(resource => resource.values.every(value => value.mime_type !== rootGetters['validation/textMimeType']))
```
bulatgab, line 366: 👏

src/store/flow/resource.ts
bulatgab: Thanks a lot for this change, to me it looks easier to understand! 🤗

bulatgab (approved):

PR #338, opened 2022-09-06
VMO-6792 Detecting changes in Flows

object-required:
- Block movement is now a change
- Block toolbar visibility is not


bulatgab:
(not related to your changes 😉)

Maybe it's not critical for the top-level component like `InteractionDesigner` but we could use `unsubscribe`:
```
  created() {
    this.unsubscribe = this.$store.subscribe(...)
  },
  beforeDestroy() {
    this.unsubscribe();
  }
```

src/views/InteractionDesigner.vue
```diff
@@ -314,13 +314,26 @@ export class InteractionDesigner extends mixins(Lang, Routes) {
...
+
+    // Visibility of block toolbar is not a flow change
+    if (type === 'flow/block_updateVendorMetadataByPath' && payload.path === 'floip.ui_metadata.should_show_block_tool_bar') {
+      return
+    }
+
+    // Only consider flow mutations
+    if (type.startsWith('flow/') === false) {
       return
     }
```
bulatgab, line 339: I'm worried that it's getting too specific (and therefore fragile) 🤔 Do you agree?

I wonder if we could compare the old flow container with the new one instead, something like `_.isEqual(oldFlowContainer, newFlowContainer)` which would probably make it more robust?
We would have to ignore the `container.isCreated` because according to `flow_persistImport` it's not being sent.

The problem is that it would be expensive to run on every change but we could simply `debounce`.
Another problem is that `subscribe` doesn't give us the old state. I don't have a solution for that yet...

bulatgab: Now that I think of it, debouncing a function that has these parameters `(oldState, newState) => {...}` probably would't work anyway.

What if we were saving the flow state somewhere every time we were successfully persisting the flow?
And what if we were comparing that saved flowContainer to the new ones we receive?

P.S. Also, related is the idea of undo/redo buttons (which hasn't been discussed with anyone though). For that we could probably push the new flowContainers into some array and reuse them if 'undo' is pressed.

object-required: We do have `hasChanges` in trees which utilizes `JSON.stringify` approach. It has a couple of TODOs though, so I'd suggest for us to create a tech debt ticket to refactor changes detection later.

bulatgab: Interesting, thanks!
Approved.

bulatgab (approved):

src/views/InteractionDesigner.vue
bulatgab: https://viamoinc.atlassian.net/browse/VMO-6926

PR #344, opened 2022-09-22
VMO-7069 Escape choices in Select One Response

object-required:
This PR adds escaping to the `choicesToExpression` method. The UI uses `vendor_metadata` to display selected options, so we don't need a reverse transformation.


safydy-r (approved):
lgtm

bulatgab (approved):

PR #347, opened 2022-09-23
Feature/vmo 6541/emit d ts files

bulatgab:
## TL;DR
Types are here!
Use `yarn build-bundle` as before, it will additionally generate typescript declaration files under `dist/types`.  

## Commentary
1. We have been using `vue-cli-service` for bundling. However, it doesn't seem to work for emitting `.d.ts` typescript declaration files. Therefore we need to use an additional `vue-tsc` library for that. We are still using `vue-cli-service` for everything as before but `yarn build-bundle` now additionally calls `vue-tsc`.
2. `vue-tsc` can be configured the same way as `tsc`, either using command-line options or via a tsconfig file. We need to specify `include` option to filter out unit tests and storybook files and command-line options don't support `include`. Therefore we have a new `vue-tsc.tsconfig.json` file. Yes, another config file 😔.
3. Another little fix - `vue-tsc` keeps the path aliases as is, e.g. `import ... from '@/store'`. This produces an error when used in the embedding app. We need to call another library `tsc-alias` to resolve the aliases, e.g. into `import ... from '../../../store'`
4. I haven't merged the `.d.ts` files into one so far, I placed them under `dist/types` and kept the structure of `src` directory. I've seen other projects to use unmerged files too, but yes, It would be cleaner to have just one `.d.ts` in the future.
5. Errors. I'm getting 9 errors right now at the `vue-tsc` stage, mostly in dependencies. It doesn't seem to break the process. Let me know if this should be fixed. Occasionally I'm getting a 'loader' error on the first `vue-cli-service` step which fails the whole thing. This goes away if the command is rerun. I wonder if it's a memory thing.

For the failed attempt to use webpack for this purpose see this PR: https://github.com/FLOIP/flow-builder/pull/353


bulatgab:
@AndrewsGomesViamo I'm adding new dependencies here and Safydy recommended to add you as a reviewer 🙂

safydy-r (approved):
Thanks Bulat, good job here.

object-required (approved):

AndrewsGomesViamo (approved):
LGTM

I'm not seen anything that may become a blocker to some future migration to vue 3

pascaleli (approved):

PR #348, opened 2022-09-23
VMO-7040 handle validations & update styles

safydy-r:
- [x] update destination branch once vmo-7039 is merged

[Overview]
Screen Shot 2022-09-23 at 3 23 55 PM


src/components/interaction-designer/BuilderCanvas.vue
```diff
@@ -31,45 +31,13 @@ const MARGIN_HEIGHT_CORRECTION_PX = -10
...
 const DEBOUNCE_SCROLL_TIMER_MS = 300
-const DEBOUNCE_VALIDATION_TIMER_MS = 300
+// const DEBOUNCE_VALIDATION_TIMER_MS = 300
 
 @Component({})
 export class BuilderCanvas extends Vue {
   @Prop() block!: IBlock
   @Prop({default: 0}) widthAdjustment!: number
 
-  // ###### Validation API Watchers [
```
safydy-r, line 41: Moved these into InteractionDesigner.vue so any changes under that component will call the validation, meaning changes in BuilderCanvas.vue and in ResourceViewer.vue

src/components/interaction-designer/block-editors/SimplifiedNameEditor.vue
```diff
@@ -0,0 +1,94 @@
+<template>
```
safydy-r, line 1: I created this new component instead of using the existing `NameEditor.vue` because there are many differences between the 02 desired UI. With this approach, it will be cleaner.

src/views/InteractionDesigner.vue
```diff
@@ -146,6 +170,27 @@ export class InteractionDesigner extends mixins(Lang, Routes) {
   @builderNamespace.Getter isResourceViewerCanvasEnabled!: boolean
   @clipboardNamespace.Getter isSimulatorActive!: boolean
 
+  // ###### Validation API Watchers [
```
safydy-r, line 173: - [ ] re-arrange this comment to be at the beginning

src/components/interaction-designer/block-editors/index.ts
```diff
@@ -20,6 +20,7 @@ export * from './MaximumNumericEditor.vue'
 export * from './MinimumChoicesEditor.vue'
 export * from './MinimumNumericEditor.vue'
 export * from './NameEditor.vue'
+export * from './SimplifiedNameEditor.vue'
```
bulatgab, line 23: minor: let's put it after 'SemanticLabelEditor' to keep alphabetical order :)

src/components/interaction-designer/block-editors/SimplifiedNameEditor.vue
```diff
@@ -0,0 +1,94 @@
...
+  updated() {
+    if (this.blockName === '') {
+      this.editBlockName = true
+    }
+  }
+
+  filterName(e: KeyboardEvent): void {
+    this.activateEditing()
+
+    // Consider user input as name editing
```
bulatgab, line 68: This comment was referring to `activateEditing` line (in the `NameEditor`), let's move it up, or better remove it altogether?

Also, can we add this to the method itself:  `// allow [a-zA-Z0-9_] only`?
Maybe it's just me but I didn't understand at first glance why we were using this _along with_ the regular v-model.

src/components/interaction-designer/block-editors/index.ts
safydy-r: thnx, addressed in this PR: https://github.com/FLOIP/flow-builder/pull/350/files#diff-57955344a9992dbe73e7ffbeb0d008abf5ad25d5049f7cac11daa226beb252a9R26

src/components/interaction-designer/block-editors/SimplifiedNameEditor.vue
safydy-r: @bulatgab , I think the current behaviour does only accept `[a-zA-Z0-9_]` with the below  `e.key.match(/\W+|Enter/g)`, or did I misunderstand your proposal ?

`filterName` is used for `@keydown=`, could you explain a bit the 2nd question?

safydy-r: ✅ comment cleaned in the recent subtask PR: https://github.com/FLOIP/flow-builder/pull/350/files#diff-1635390784b6b76ab978432bd9c5593ebe899323e62e83332ac0c64f3be6a029L68 (I made it there to avoid cascading updates)

object-required (approved):

src/views/InteractionDesigner.vue
```diff
@@ -146,6 +170,27 @@ export class InteractionDesigner extends mixins(Lang, Routes) {
...
+  @Watch('activeFlow.resources', {deep: true, immediate: true})
+  async onResourcesOnActiveFlowChanged(newResources: IResources, oldResources: IResources): Promise<void> {
+    console.debug('watch/activeFlow.resources:', 'resources inside active flow have changed', `from ${this.mainComponent}.`, 'Validating ...')
+    await this.validate_resourcesOnSupportedValues({
+      resources: newResources,
+      supportedModes: this.activeFlow.supported_modes,
+      supportedLanguages: this.activeFlow.languages,
+    })
+  }
+
```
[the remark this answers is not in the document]
safydy-r: Thanks @bulatgab for this initiative, I'd be happy to see how this would go in store ;)
Could you pls create a new ticket for that under the resource-view epic https://viamoinc.atlassian.net/browse/VMO-6150 ? (because a subtask in phase1 might not be appropriate, in case we cannot finish that new refractor in time)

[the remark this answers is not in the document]
bulatgab: You're right, I shouldn't have dived into it right away. I'll create a new ticket for that.
Thanks for approving the suggested change 👍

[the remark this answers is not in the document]
bulatgab: https://viamoinc.atlassian.net/browse/VMO-7115

src/components/interaction-designer/block-editors/SimplifiedNameEditor.vue
safydy-r: comment added as well ;)

PR #350, opened 2022-09-27
Feature/vmo 6851/vmo 7095/view mode on block components

safydy-r:
[overview]
Screen Shot 2022-09-26 at 6 12 37 PM


src/components/interaction-designer/resource-viewer/ResourceViewerBlock.vue
```diff
@@ -5,15 +5,17 @@
       <div class="ml-auto"><h6>{{trans(`flow-builder.${block.type}`).toUpperCase()}}</h6></div>
     </div>
     <div class="block-content">
-      <div class="d-flex">
-        <simplified-name-editor :block="block" class="col-3 p-0"/>
-        <div v-if="!hasContent" class="col-3">
-          <div class="badge badge-info">{{trans('flow-builder.no-content-block-info')}}</div>
+      <fieldset :disabled="!isEditable">
```
safydy-r, line 8: FYI @bulatgab , with this we will be able to disable anything inside. But the pbm is, the btn "show/hide" will be disabled as well. I will leave this as it is for now, and will put a new AC in your ticket https://viamoinc.atlassian.net/browse/VMO-7101 to handle that

bulatgab: Thank you for noticing and bringing this up beforehand 👍  and for posting a helpful a link for the problem

PR #355, opened 2022-09-30
VMO-7087 Rename Set Contact Property

object-required:


object-required:
Closing the PR because we chose a different approach for the (VMO-7087)[https://viamoinc.atlassian.net/browse/VMO-7087] story.

PR #357, opened 2022-10-04
VMO-7043 attempt to implement from watcher

safydy-r:


src/views/InteractionDesigner.vue
```diff
@@ -311,6 +325,35 @@ export class InteractionDesigner extends mixins(Lang, Routes) {
   @builderNamespace.Mutation setInteractionDesignerBoundingClientRect!: (value: DOMRect) => void
   @builderNamespace.Action setIsEditable!: (arg0: boolean) => void
   @builderNamespace.Action setHasFlowChanges!: (arg0: boolean) => void
+
+  created(): void {
```
safydy-r, line 329: phpStorm re-arranged this

safydy-r:
declining this because we have a big pbm with watchers here:
- every time we persist the flow, everything gets slightly changed from backend (eg: resources list got cleaned for some reason), that is caught by the watcher again, and it triggers another persistence >> It was quite difficult to fix.

PR #366, opened 2022-10-11
Feature/vmo 6851/vmo 7195/fix missing resource variant

safydy-r:
The flow persistence is cleaning the resource, see: https://github.com/FLOIP/flow-builder/blob/master/src/store/flow/flow.ts#L210

but for some release we don't have this `flow_addMissingResourceValues` from @object-required anymore, very strange.
The solution:
- put flow_addMissingResourceValues back
-  and call it when we edit the flow languages/modes

TODO after merge
- [ ] udpate voto5

### DIFF
--- a/src/store/flow/flow.ts
+++ b/src/store/flow/flow.ts
@@ -234,7 +234,6 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
     }
     try {
       const {data} = await axios[restVerb](persistRoute, omit(cleanupFlowResources(flowContainer, rootGetters['validation/choiceMimeType']), ['isCreated']))
-      console.debug('test', data)
       commit('flow_setFlowContainer', data)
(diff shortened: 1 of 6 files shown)


bulatgab (approved):

src/store/flow/flow.ts
```diff
@@ -406,6 +406,41 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
     })
   },
 
+  flow_addMissingResourceValues({getters, rootGetters, dispatch}): void {
```
safydy-r, line 409: @object-required , any idea about why your [change](https://github.com/FLOIP/flow-builder/pull/335/files#diff-eea673c657b7edcdbf00ab176fcc8ec0a84d21e8a8dfa9ff8a014c7510a144a0R358) is not in `master` anymore

PR #368, opened 2022-10-13
VMO-7146 Remove default exit label overrides

object-required:
See screenshots in the [Voto5 PR #2793](https://bitbucket.org/voto/voto5/pull-requests/2793/vmo-7146-default-exit-names).

Screen Shot 2022-10-18 at 1 31 49 PM


builder.config.json
```diff
@@ -200,9 +194,6 @@
           "primary": {
             "name": "Valid",
             "test": "@ISNUMBER(block.value)"
-          },
-          "default": {
-            "name": "Invalid"
```
safydy-r, line 205: @object-required , As discussed, I'll put my review on hold until we have a clarification from Nancy what she mean about 
-`The final exit for every block type (in every mode) should be named “Default”`
- `This is true for most block right now, but should be updated for`
 (especially for Numeric block)

bulatgab (approved):
LGTM 👍

builder.config.json
safydy-r: This is resolved after our standup meeting

safydy-r (approved):
I think this is good now, thanks

PR #370, opened 2022-10-17
VMO-7198 autolink logic

bulatgab:
Simple cleanup of all autolink logic (already dead and unused, so mainly - to avoid any confusion and bugs) because it's now moved to the vendor side.


safydy-r (approved):
Thanks for this clean up

PR #374, opened 2022-10-25
VMO-7198 autolink logic with resource helpers

bulatgab:
I needed to use a helper function on the voto5 side but it wasn't exported.
So in this simple PR:
- helpers functions are moved from `store/flow/resource` to `store/flow/utils/resourceHelpers`
- in `lib.ts` the helpers are exported
- in other files the imports are updated


object-required (approved):

safydy-r (approved):
LGTM, could you pls confirm the `yarn serve` has no TS errors ?

PR #376, opened 2022-11-02
Feature/vmo 6868/vmo 7272/add incomplete flag and filter

safydy-r:
we needed this to:
- add extra information on the resource validations, which we would use in the vendor repo
- add needed slots on vue component


src/store/flow/utils/resourceHelpers.ts
```diff
@@ -165,3 +165,14 @@ export function findBlockRelatedResourcesUuids({block}: {block: IBlock}): IResou
...
+/**
+ * Compute resource index (cell index) for a table having X languages and Y modes
+ *
+ * @param languageIndex
+ * @param modeIndex
+ * @param supportedModeLength
+ */
+export function computeResourceIndex(languageIndex: number, modeIndex: number, supportedModeLength: number): number {
+  return languageIndex * supportedModeLength + modeIndex
+}
```
object-required, line 178: I'd like to suggest renaming `supportedModeLength` for extra clarity. 
For myself, it sounds like `'USSD'.length === 4` 😃 

```suggestion
export function computeResourceIndex(languageIndex: number, modeIndex: number, suppoertedModesCount: number): number {
  return languageIndex * suppoertedModesCount + modeIndex
}
```

src/store/validation/index.ts
```diff
@@ -33,12 +34,23 @@ export interface IValidationStatusContext {
...
 
+/**
+ * Is valid by mode index OR lang index:
+ * - if 1st key is mode index, then the array values are lang index
+ * - if the 1st key is lang index, then the array values are mode index
+ */
+export interface IInvalidResourcesCounterBy {
+  modeIndex?: {[key: number]: number},
+  langIndex?: {[key: number]: number},
+}
```
object-required, line 45: Could you confirm my understanding, please?
An object of `IInvalidResourcesCounterBy` will be shaped as follows:

```js
{
  modeIndex: {
    "0": 12 // there are 12 invalid resources for mode with index "0"
  },
  langIndex: {
    "1": 34 // there are 34 invalid resources for language with index "1"
  }
}
```

If that's the case, we may rename the interface as

```suggestion
/**
 * Number of invalid resources either by mode index or lang index.
 */
export interface IInvalidResourceCounters {
  byModeIndex?: {[key: number]: number},
  byLangIndex?: {[key: number]: number},
}
```

src/store/validation/index.ts
```diff
@@ -316,6 +329,40 @@ export const validationActions: ActionTree<IValidationState, IRootState> = {
...
+    errors?.forEach((error, index) => {
+      // replace non-numeric characters from dataPath (which could be something like `values/1/value`)
+      const currentResourceIndex = error.dataPath.replace(/\D/g, '')
+      modes.forEach((mode: SupportedMode, modeIndex: number) => {
+        languages.forEach((lang: ILanguage, langIndex: number) => {
+          const resourceIndex: string = computeResourceIndex(langIndex, modeIndex, modes.length).toString()
+          // if resourceIndex === currentResourceIndex, isValid === true, false otherwise
+          if (resourceIndex === currentResourceIndex) {
+            set(invalidCounterByMode, modeIndex, get(invalidCounterByMode, modeIndex, 0) + 1)
+            set(invalidCounterByLang, langIndex, get(invalidCounterByLang, langIndex, 0) + 1)
```
object-required, line 356: I found [`_.update`](https://devdocs.io/lodash~4/index#update) provides a neat way to increment the counter:

```js
update(invalidCounterByMode, modeIndex, n => n ?? 0 + 1)
```

object-required (approved):
Thank you, @safydy-r for great work here. I added a couple of suggestions that may or may not make code easier to understand 😉 

I also wondered if we should use `iso_639_3` language codes and actual mode names. This could make runtime debugging with Vue DevTools more convenient, but the tradeoff is code complexity.

src/store/flow/utils/resourceHelpers.ts
safydy-r: Got you, thanks. Renamed it in a different commit as this suggestion has a type ;)

src/store/validation/index.ts
safydy-r: you're correct, I will add this example from you to be part of the JS-Doc, thanks ;)

safydy-r: and updated the `then the array values are lang index` because that is not true

safydy-r:
@object-required , thanks for the great feedback on this PR 💯

PR #379, opened 2022-11-08
Feature/vmo 6880/vmo 7348/tag selector per block

bulatgab:
voto5 PR https://bitbucket.org/voto/voto5/pull-requests/2864/feature-vmo-6868-vmo-7348-tag-selector-per
(the demo can be found there)


src/components/interaction-designer/block-editors/TagSelector.vue
```diff
@@ -17,7 +17,17 @@
         :close-on-select="false"
         :taggable="taggable"
         :tag-placeholder="taggable ? trans('flow-builder.create-a-tag-prompt') : ''"
-        @tag="addTag" />
+        @tag="addTag">
+        <template #clear>
+          <font-awesome-icon
+            v-if="selectedTags.length"
+            v-b-tooltip.hover="trans('flow-builder.clear-all')"
+            style="position: absolute; right: 41px; top: 13px; cursor: pointer"
```
bulatgab, line 25: Extracting this to a class didn't work. Probably, a class becomes unavailable when passed via the slot.

bulatgab: also, the 'clear' slot doesn't work well. Vue-multiselect uses 'position: absolute' here and there, its layout is not great, so we need to use 'position: absolute' too.

bulatgab: Without it the `x` goes on top
Screen Shot 2022-11-07 at 11 44 59 PM

bulatgab: Never mind. I've used the `>>>` deep combinator and it worked

dist/types/store/resource-viewer/index.d.ts
```diff
@@ -0,0 +1,18 @@
...
+import { IRootState } from '../../store';
+import { IBlock } from '@floip/flow-runner';
+export interface IResourceViewerState {
+    /**
+     * Block uuids to hide,
+     * Will be used for resource-viewer component, so the consumer repo can set a custom filters and override this state if needed
+     *
+     * By default, it's empty, meaning there is no filter at all
+     */
+    blockUuidsToHide: IBlock['uuid'][];
```
safydy-r, line 11: @bulatgab , I'm curious why do we need those changes in current PR ?

bulatgab: I think it is not new, it was just missing.
Probably after creating the `index.ts` source file and running `yarn build-bundle` this `.d.ts` was generated but not git-added.
I think we could maybe append `git add dist/` to the `"build-bundle"` script.

safydy-r: Got you, I'd didn't notice it was in `dist/`. thanks

safydy-r (approved):
LGTM

object-required (approved):

PR #382, opened 2022-11-14
Feature/vmo 6880/vmo 7351/search frontend logic

bulatgab:
This change re-exposes the `tag` slot (with its props) in the TagSelector.
It allows us to restyle the tags on the voto5 side and highlight those matching the search query.


object-required (approved):

safydy-r (approved):
lgtm

PR #383, opened 2022-11-15
VMO-7349 Resource view filers

object-required:


src/components/interaction-designer/ResourceViewer.vue
```diff
@@ -23,10 +23,21 @@ export const DEBOUNCE_FLOW_PERSIST_MS = 1500
...
-    ...mapGetters('resource-viewer', [
-      'filteredBlocks',
-    ]),
+    ...mapGetters({
+      activeFlow: 'flow/activeFlow',
+    }),
+    visibleBlocks() {
+      return Array.isArray(this.filteredBlocks)
+        ? this.filteredBlocks
+        : this.activeFlow?.blocks ?? []
```
object-required, line 39: Resource Viewer will now display `filterdBlocks` from props if present, and all blocks from the `activeFlow` otherwise.

bulatgab: `visibleBlocks` - good name 👍

src/components/interaction-designer/ResourceViewer.vue
```diff
@@ -23,10 +23,21 @@ export const DEBOUNCE_FLOW_PERSIST_MS = 1500
 export default {
   name: 'ResourceViewer',
   mixins: [Lang],
+  props: {
+    filteredBlocks: {
+      type: Array,
+      default: () => null,
```
bulatgab, line 29: is `default` needed?

object-required: I believe it's either required or some default value, null seemed reasonable for an array

bulatgab: Sorry, I see now that we have an ESLint rule requiring a default value for everything that's not marked as required.

bulatgab: The value of `() => null` is an interesting choice.
Have you considered using `null/undefined`?
I think the visibleBlocks could then use nullish coalescing like
`return this.filteredBlocks ?? this.activeFlow?.blocks ?? []`.

bulatgab (approved):

safydy-r (approved):
Nice one, thanx 💯

PR #385, opened 2022-11-18
Feature/vmo 7376/fix resource indexes

bulatgab:
## The problem
1. The flow spec doesn't enforce `flow.languages`, `flow.supported_modes`, `resource.values` to be sorted.
So for example, if we are importing a flow, these things are not guaranteed to be sorted.
2. Additionally, these arrays can change in various ways - the user can add and remove languages and modes

## Proposed solution
Therefore, we should not rely on their order:
- we should avoid using indexes of languages, modes, and resource values
- there is no point in sorting these arrays when we are creating or modifying them

Instead, we should use IDs of these items if we want to find them in their arrays:
- language - `language.id`
- mode - `mode` itself (a string)
- resource value - `resourceValue.language_id` + `resourceValue.modes` combination.

## Note
The difficult point is the validation by `ajv`. It outputs errors for the resource values using their indexes - the error path might look like: `values/1/value`.
I saw two options (and I chose the second one).
- Option 1
We could, again, use mode and language IDs and modify the error paths to something like `values/SMS/eng/value`.
However, the backend would have to update its validation as well.
(We need consistency because if both FE and BE find an error at the same path, we might want to remove duplicates - which we are not doing at the moment).
- Option 2
I have ended up keeping the `values/1/value` paths. And in the `ResourceEditorCell` I added converting language and mode into a resource value index using `findIndexForResourceVariant`.


src/components/interaction-designer/flow-editors/FlowEditor.vue
```diff
@@ -98,59 +92,8 @@ export class FlowEditor extends mixins(Lang) {
...
-    // flow modes/languages could have been changed, so
-    // 1. Trigger validation for all blocks
-    await this.validate_allBlocksWithinFlow()
-    // 2. Trigger validation for resources
-    await this.validate_resourcesOnSupportedValues({
-      resources: this.activeFlow.resources,
-      supportedModes: this.activeFlow.supported_modes,
-      supportedLanguages: this.activeFlow.languages,
-    })
-  }
```
bulatgab, line 112: moved to store

src/components/interaction-designer/flow-editors/ModesEditor.vue
```diff
@@ -15,10 +15,12 @@
 import VueMultiselect from 'vue-multiselect'
 import {Component, Prop} from 'vue-property-decorator'
 import {IFlow, SupportedMode} from '@floip/flow-runner'
-
 import Lang from '@/lib/filters/lang'
 import {mixins} from 'vue-class-component'
-import {Getter} from 'vuex-class'
+import {namespace, State} from 'vuex-class'
+import {orderModes} from '@/store/flow/flow'
```
bulatgab, line 21: `orderModes` - modes are stored in unsorted form, but we sort them here - on the surface, in the UI - to display in certain order

src/components/interaction-designer/resource-editors/PerLanguageResourceEditor.vue
```diff
@@ -22,17 +21,15 @@
...
-} from '@floip/flow-runner'
+import {IBlock, IFlow, ILanguage} from '@floip/flow-runner'
 import Lang from '@/lib/filters/lang'
 import Permissions from '@/lib/mixins/Permissions'
 import Routes from '@/lib/mixins/Routes'
 import FlowUploader from '@/lib/mixins/FlowUploader'
 import {Component, Prop} from 'vue-property-decorator'
 import {mixins} from 'vue-class-component'
 import {TabsPlugin} from 'bootstrap-vue'
+import {orderLanguages} from '@/store/flow/flow'
```
bulatgab, line 32: again, the `flow.languages` are unsorted (and we should not rely on its order). But we are sorting them here - in the UI - to display in certain order

src/store/flow/flow.ts
```diff
@@ -16,19 +16,30 @@ import {IdGeneratorUuidV4} from '@floip/flow-runner/dist/domain/IdGeneratorUuidV
...
 import {ErrorObject} from 'ajv'
 import {removeFlowValueByPath, updateFlowValueByPath} from '@/store/flow/utils/vuexBlockAndFlowHelpers'
 import {ConfigFieldType} from '@/store/flow/block'
 import {IFlowsState} from '.'
 import {mergeFlowContainer} from './utils/importHelpers'
 
-export function orderLanguages(LanguagesList: ILanguage[]): ILanguage[] {
-  return sortBy(LanguagesList, ['label'])
-}
-
```
bulatgab, line 31: move to the end of file

src/store/flow/flow.ts
```diff
@@ -81,28 +92,6 @@ export const getters: GetterTree<IFlowsState, IRootState> = {
...
-    const order = [
-      SupportedMode.IVR,
-      SupportedMode.SMS,
-      SupportedMode.USSD,
-      SupportedMode.TEXT,
-      SupportedMode.RICH_MESSAGING,
-      SupportedMode.OFFLINE,
-    ]
-    return sortBy(getters.activeFlow.supported_modes, (item: SupportedMode) => order.indexOf(item))
-  },
```
bulatgab, line 105: move `orderedSupportedModes` to the end of file

src/store/flow/flow.ts
```diff
@@ -394,32 +381,29 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
-    // Let's create all languages, we might need them but if they are switched off they just don't get used
-    // It's important to use the same order to generate resource, that will be helpful for resource validation indexes
-    const values: IResourceValue[] = orderLanguages(rootState.trees.ui.languages as ILanguage[])
-      .reduce((memo: IResourceValue[], language: ILanguage) => {
-      // Let's just create all the modes, we might need them but if they are switched off they just don't get used
-      Object.values(SupportedMode).forEach((mode: SupportedMode) => {
+  async flow_createBlankResource({getters, dispatch}): Promise<IResource> {
+    const activeFlow: IFlow = getters.activeFlow
+    const modes = activeFlow.supported_modes
+    const languageIds = activeFlow.languages.map(language => language.id)
```
bulatgab, line 392: add variants only for flow langs and modes (not all available langs and modes)

src/components/interaction-designer/flow-editors/FlowEditor.vue
safydy-r: Nice, thanks

src/components/interaction-designer/resource-editors/PerLanguageResourceEditor.vue
safydy-r: Can we keep this comment in the code, I feel like that would be useful in the future?

src/store/flow/flow.ts
```diff
@@ -542,4 +545,53 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
+      removedLanguageIds: difference(oldLanguages, newLanguages).map(lang => lang.id),
+    })
+    await dispatch('validation/validate_allBlocksWithinFlow', null, {root: true})
+    await dispatch('validation/validate_resourcesOnSupportedValues', {
+      resources: flow.resources,
+      supportedModes: flow.supported_modes,
+      supportedLanguages: flow.languages,
+    }, {root: true})
+  },
+}
```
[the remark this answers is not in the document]
safydy-r: I can see from here how complex this refactor would be, thanks for the huge focus you put on this one.

src/components/interaction-designer/resource-editors/PerLanguageResourceEditor.vue
bulatgab: Sure, I will add it 👍

bulatgab: Added here now: https://github.com/FLOIP/flow-builder/pull/385/commits/3b2162286327bb03f8cecdfd9ca0e1266e62cc2c
(I didn't rebuild the project because it's only a comment, it will get picked up at the next rebuild)

PR #388, opened 2022-11-22
Feature/vmo 7477/yarn3 upgrade

AlteaDown:
Set Yarn version to 3 (berry)

Updated .gitignore && yarn.lock. Moved flow-2.0.0.js to it's own directory, and created a package.json (which is required for yarn to use a local file dependency).


src/lib/flow/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "main" : "flow-2.0.0.js",
```
safydy-r, line 2: Thanks for this

safydy-r (approved):
LGTM, 
- I consumed the latest flow-runner from npm
- tested `yarn serve` and was able to load correctly the app

PR #389, opened 2022-11-22
VMO-6880 Resource Viewer, Phase 4

object-required:


src/components/interaction-designer/ResourceViewer.vue
```diff
@@ -23,10 +24,16 @@ export const DEBOUNCE_FLOW_PERSIST_MS = 1500
 export default {
   name: 'ResourceViewer',
   mixins: [Lang],
+  props: {
+    visibleBlocks: {
+      type: Array,
+      default: () => null,
+    },
```
object-required, line 31: Hiding blocks is necessary for keeping the state of resource accordions.

src/components/interaction-designer/ResourceViewer.vue
```diff
@@ -3,7 +3,8 @@
     <slot name="vendor-header" />
     <div class="resource-viewer-contents">
       <resource-viewer-block
-        v-for="block in filteredBlocks"
+        v-for="block in this.activeFlow?.blocks ?? []"
```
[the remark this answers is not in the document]
bulatgab: this can be omitted

src/components/interaction-designer/ResourceViewer.vue
bulatgab: Could you elaborate a bit - why a list of UUIDs instead of actual blocks?
(I agree that it is better but I can't articulate)

src/components/interaction-designer/resource-editors/PerModeResourceEditor.vue
```diff
@@ -84,7 +85,38 @@ export class PerModeResourceEditor extends mixins(FlowUploader, Permissions, Rou
     [SupportedMode.OFFLINE, ['fas', 'mobile-alt']],
   ])
 
-  isPanelExpanded = {}
+  expandedPanels = {}
+  expandedPanelsSearch = {}
+
+  get hasOpenPanelsOverriden() {
```
bulatgab, line 91: Overriden -> Overridden

src/components/interaction-designer/ResourceViewer.vue
object-required: UUIDs are used to avoid possible confusion, as this is a list of block IDs not to display.

The blocks now hidden `v-show` instead of conditional `v-if` rendering. This allows persisting the state of accordions, which are nested components – otherwise they would be recreated and the information about collapsed/expanded lost.

src/components/interaction-designer/resource-editors/PerModeResourceEditor.vue
```diff
@@ -84,7 +85,38 @@ export class PerModeResourceEditor extends mixins(FlowUploader, Permissions, Rou
...
+  set isPanelExpanded(value) {
+    if (this.hasOpenPanelsOverriden) {
+      this.expandedPanelsSearch = value
+    } else {
+      this.expandedPanels = value
+    }
+  }
+
+  @Watch('openResources')
+  overrideOpenResources(resources: string[]) {
```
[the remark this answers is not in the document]
bulatgab: I'm not a fan of watchers but this seems to be the best way of updating the state when the prop changes :(
(we can't use a computed property, can we?)

src/components/interaction-designer/resource-editors/PerModeResourceEditor.vue
```diff
@@ -84,7 +85,38 @@ export class PerModeResourceEditor extends mixins(FlowUploader, Permissions, Rou
...
+      : this.expandedPanels
+  }
+
+  set isPanelExpanded(value) {
+    if (this.hasOpenPanelsOverriden) {
+      this.expandedPanelsSearch = value
+    } else {
+      this.expandedPanels = value
+    }
+  }
```
bulatgab, line 107: this setter might be removed if we modify the `updateIsPanelExpanded` a bit.
please see a proposal here:
https://github.com/FLOIP/flow-builder/commit/6e05d98c12187a42c47ede902ea89bf40f5b3e8c
(cherry-pick if you agree)
There are some types as well. And I would propose renaming expandedPanelsSearch -> expandedPanelsOverridden since we are agnostic of the search in this component and we are already using the 'overridden' term

bulatgab: ~~agnostic~~ unaware

bulatgab (approved):
Looks good to me 👍
(some optional refactoring can be done but leaving it up to you)

src/components/interaction-designer/resource-editors/PerModeResourceEditor.vue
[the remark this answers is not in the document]
object-required: Unfortunately, not in this case. The component depends on Voto5 data, and watching the prop seems to be the only way to react to change.

src/components/interaction-designer/ResourceViewer.vue
bulatgab, line 6: this can be omitted
(reposting for better visibility)

object-required:
```suggestion
        v-for="block in this.activeFlow?.blocks"
```

bulatgab: Sorry about the confusion! I meant `this`
P.S. When do I stop being such a nitpicker?

PR #392, opened 2022-11-29
Feature/vmo 7383/vmo 7521/fix duplicate keys detected

bulatgab:


safydy-r (approved):
I'd approve this, thanks for find a solution here

object-required (approved):

safydy-r:
We've decided to decline this as the ts-generator cannot handle the UNION type nor the `anyOf` approach

PR #397, opened 2022-12-06
Feature/vmo 7383/minor fixes

bulatgab:
- remove `orgTimezone` to avoid confusion. Why we can do it:
  - it doesn't contain correct value
  - we are not using it in the community builder anyway

- add `skipLibCheck` to remove console errors from npm dependencies when running `vue-tsc`. Why:
  - we don't control these libs and we can't fix them so there is no point in seeing the errors


object-required (commented):
Looks good, thank you for working on the build 💯

PR #399, opened 2022-12-07
VMO-7508 Allow updating `vendor_metadata` on block exits

object-required:


src/store/flow/block.ts
```diff
@@ -90,6 +90,11 @@ export const mutations: MutationTree<IFlowsState> = {
     const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
     block.exits = reject(block.exits, (e) => e.uuid === exit.uuid)
   },
+  block_updateExitVendorMetadata(state, {blockId, exitId, vendorMetadata}: {blockId: string, exitId: string, vendorMetadata: object}) {
+    const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
+    const exit = findBlockExitWith(exitId, block)
+    exit.vendor_metadata = vendorMetadata
+  },
```
object-required, line 97: This mutation is used in Voto5 to update exit conditions.

src/store/flow/block-types/BaseBlockStore.ts
```diff
@@ -46,11 +47,9 @@ export const actions = {
...
         },
       },
-    }, {
-      // Extra vendor_metadata from consumer side (eg: some configs under a new namespace)
-      vendor_metadata: await dispatch('initiateExtraVendorConfig'),
-    },
-    )
+    }
+
+    const mainProps = defaultsDeep({}, vendorMetadata, props, defaultProps)
```
object-required, line 52: `_.defaults` method [https://devdocs.io/lodash~4/index#defaults](won't override) `vendorMetadata`, so I changed the order of assignments.

src/store/flow/block.ts
safydy: Thanks for this, I was thinking to add the same thing

safydy (approved):
LGTM

PR #404, opened 2022-12-09
CORE-67 Allow appending exits

object-required:
`block_addExit` inserts exits before the last element of `block.exits[]`, and we need it to append exits at the end instead. This PR introduces an optional parameter `append` to support this behaviour.

### DIFF
--- a/src/store/flow/block.ts
+++ b/src/store/flow/block.ts
@@ -80,10 +80,14 @@ export const mutations: MutationTree<IFlowsState> = {
     const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
     findBlockExitWith(exitId, block).semantic_label = value
   },
-  block_addExit(state, {blockId, exit}: {blockId: string, exit: IBlockExit}) {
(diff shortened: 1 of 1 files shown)


src/store/flow/block.ts
```diff
@@ -80,10 +80,14 @@ export const mutations: MutationTree<IFlowsState> = {
     const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
     findBlockExitWith(exitId, block).semantic_label = value
   },
-  block_addExit(state, {blockId, exit}: {blockId: string, exit: IBlockExit}) {
+  block_addExit(state, {blockId, exit, append}: {blockId: string, exit: IBlockExit, append?: boolean}) {
```
safydy-r, line 83: @object-required , that param is a boolean, let's follow the boolean naming convention with prefix `has`, `should`, `is`, etc
Maybe `shouldAppend` is good ?

safydy-r (approved):
Pre-approved with change request

PR #405, opened 2022-12-12
VMO-7508 Branching Block Integration (CORE-44)

object-required:


src/store/flow/block.ts
```diff
@@ -80,16 +80,31 @@ export const mutations: MutationTree<IFlowsState> = {
...
-  block_addExit(state, {blockId, exit}: {blockId: string, exit: IBlockExit}) {
+  block_addExit(state, {blockId, exit, shouldAppend}: {blockId: string, exit: IBlockExit, shouldAppend?: boolean}) {
     const block = findBlockOnActiveFlowWith(blockId, state as unknown as IContext)
-    // insert before default exit
-    block.exits.splice(block.exits.length - 1, 0, exit)
+
+    if (shouldAppend) {
+      block.exits.push(exit)
+    } else {
+      block.exits.splice(block.exits.length - 1, 0, exit)
```
bulatgab, line 89: could we please add a short comment what this is doing? is this inserting before the last element (default exit?)

bulatgab (approved):

PR #407, opened 2022-12-16
Feature/core 20/core 143/branching block group membership vue

bulatgab:


object-required (approved):
I couldn't find `GroupSelector` being used in Voto5

safydy-r (approved):

PR #410, opened 2023-01-10
Feature/core 48/customize snake case method for label generation

safydy-r:
Basically we just replace spaces with underscore and make the string to lower case.

update overview
Screen Shot 2023-01-16 at 5 56 14 PM

### DIFF
--- a/src/utils/string-utils.ts
+++ b/src/utils/string-utils.ts
@@ -1,14 +1,17 @@
-// eslint-disable-next-line import/prefer-default-export
-export function removeUnderScoreCharBeforeDigits(value: string) {
-  const charBeforeDigits = value.match(/_[0-9]/g) || []
-  let finalValue = value
-  if (charBeforeDigits === null) {
-    return value
-  } else {
-    charBeforeDigits.forEach((foundChar: string) => {
-      finalValue = finalValue.replaceAll(foundChar, foundChar.split('_')[1])
-    })
+import {snakeCase} from 'lodash'
 
-    return finalValue
+/**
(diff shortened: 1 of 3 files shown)


src/utils/string-utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (charBeforeDigits === null) {
+    return value
+  } else {
+    charBeforeDigits.forEach((foundChar: string) => {
+      finalValue = finalValue.replaceAll(foundChar, foundChar.split('_')[1])
+    })
+
+    return finalValue
+  }
+}
```
bulatgab, line 14: There is probably a way to achieve this in one 'replaceAll' with regex. Let me see...

object-required (approved):

src/utils/string-utils.ts
bulatgab:
```
snakeCase("01 a23 b45 c67c 89de").replace(/_(\d+)/g, '$1')
// output: 01_a23_b45_c67_c89_de
```

Ref: Eloquent Javascript, https://eloquentjavascript.net/09_regexp.html#h_k0YuTOu54D

safydy-r: This is what I've been looking for yesterday. Thanks so much, I will make a quick refactor

bulatgab: You're welcome :)
@object-required Could you please double-check?

bulatgab (approved):

object-required (approved):

PR #416, opened 2023-02-08
Feature/CORE-154/CORE-269/merge duplicate fe and be validations

safydy-r:
Summary:
- we ignore backend validations (for blocks, or resources) if there is a frontend validation related to the proposed `dataPath`

Eg:
Screen Shot 2023-02-08 at 4 07 36 PM


object-required (approved):

src/store/validation/validationHelpers.ts
```diff
@@ -170,20 +170,18 @@ export function getLocalizedAjvErrors(keyPrefix: string, ajvErrors?: ErrorObject
...
-export function getLocalizedBackendErrors(keyPrefix: string, blockErrors: { message: string }[]): ErrorObject[] | null {
+export function getLocalizedBackendErrors(keyPrefix: string, blockErrors: { message: string, dataPath: string }[]): ErrorObject[] | null {
   if (!Array.isArray(blockErrors)) {
     console.debug('getLocalizedBackendErrors', 'returning null as blockErrors is not an array', blockErrors)
     return null
   }
 
-  return blockErrors.map((error: { message: string }) => {
-    // replace all digital indexes to 'x' so we can easily localize them
-    const noIndexMessage = error.message.replaceAll(/\d+/g, 'x')
```
bulatgab, line 181: we are no longer replacing indexes with 'x'?

bulatgab:
Thank you for fixing this, Safydy 👍👍

src/store/validation/validationHelpers.ts
safydy-r: @bulatgab , that's true.
From now:
- the backend sends the `dataPath` so we can identify if there is a duplicate and retain only the frontend's. Before we can only identify the backend validation from the weird json schema result, which we will needed to all digital indexes to 'x'. 
- on localization side, the backend sends `message` which we will just consider as localization key

cc: @object-required

PR #418, opened 2023-02-09
Feature/CORE-264/expand click area to show the block config

safydy-r:
[result]
core-264


src/components/interaction-designer/Block.vue
```diff
@@ -629,7 +629,24 @@ export class Block extends mixins(Lang) {
   }
 
   selectBlock(): void {
-    const routerName = this.isBlockEditorOpen ? 'block-selected-details' : 'block-selected'
+    let shouldBlockEditorBeVisible
+    if (this.activeBlockId !== this.block.uuid) {
```
safydy-r, line 633: if we click a different block, then it should always be TRUE

bulatgab (approved):

yeboaasare (approved):

PR #428, opened 2023-03-10
CORE-310 add notification module

safydy-r:
Summary:
I just added a notification module to wrap the `vue-toastification` options with a `GENERAL_OPTIONS`, which will help us to just call something like this later.
```
    this.toastError({
      content: 'test error!',
      options: {
        position: 'top-center',
      },
    })
    this.toastSuccess({
      content: 'test success!',
    })
    this.toastWarning({
      content: 'test warning!',
    })
    this.toastInfo({
      content: 'test info!',
      options: {
        timeout: 5000,
      },
    })
```
These are example of action dispatches from a vue component, which give us this kind of render
core-310

**Note**: I didn't replicate the FlightMonitor component because we don't really need such level of complexity in this repo.

### DIFF
--- a/src/store/notification/index.ts
+++ b/src/store/notification/index.ts
@@ -1,0 +1,87 @@
+import Vue from 'vue'
+import {ActionTree, Module} from 'vuex'
+import {IRootState} from '@/store'
+import {ToastOptions} from 'vue-toastification/dist/types/src/types'
+import {POSITION} from 'vue-toastification/src/ts/constants'
+import {TYPE} from 'vue-toastification/dist/types/src/ts/constants'
+
+/**
+ * For more details about available options, see https://vue-toastification.maronato.dev/
+ */
+const GENERAL_OPTIONS: ToastOptions = {
+  position: POSITION.TOP_RIGHT,
+  draggable: true,
+  draggablePercent: 0.6,
+  closeOnClick: false,
+  timeout: false,
+}
+
+/**
+ * Note, we can use a custom component for `content`, not only texts
+ */
+export type IToastContent = string | object
+
+export interface IToast {
+  content: IToastContent,
+  options: ToastOptions,
+}
+
+export interface INotificationState {}
+
+export const stateFactory = (): INotificationState => ({})
+
+export const notificationActions: ActionTree<INotificationState, IRootState> = {
+  toast(_context, {content, options}: IToast) {
+    Vue.$toast(content, {
+      ...GENERAL_OPTIONS,
+      ...options,
+    } as ToastOptions)
+  },
+  toastError({dispatch}, {content, options}: IToast) {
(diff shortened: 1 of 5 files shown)


src/store/notification/index.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Note, we can use a custom component for `content`, not only texts
+ */
+export type IToastContent = string | object
+
+export interface IToast {
+  content: IToastContent,
+  options: ToastOptions,
+}
```
bulatgab, line 27: Would it be okay to move IToast _above_ IToastContent, maybe even to the very top of the file?
That way the reader would see the higher-level type first

src/store/notification/index.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+  toastError({dispatch}, {content, options}: IToast) {
+    const defaultOptions: ToastOptions = {} as ToastOptions
+    Vue.$toast.error(content, {
+      ...GENERAL_OPTIONS,
+      ...defaultOptions,
+      ...options,
+    } as (ToastOptions & {
+      type?: TYPE.ERROR | undefined
+    }) | undefined)
```
bulatgab, line 48: about the `as (ToastOptions & {type?: TYPE.ERROR | undefined}) | undefined`

I think you can also do something like this:
```
import {TYPE as ToastType} from 'vue-toastification/dist/types/src/ts/constants'

export interface IToast<T extends ToastType> {
  content: IToastContent,
  options: ToastOptions & {type: T},
}

toast(_context, {content, options}: IToast<ToastType>) {
    Vue.$toast(content, {
      ...GENERAL_OPTIONS,
      ...options,
    })
  },
  toastError({dispatch}, {content, options}: IToast<ToastType.ERROR>) {
    Vue.$toast.error(content, {
      ...GENERAL_OPTIONS,
      ...options,
    })
  },
```

src/store/notification/index.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+  toastSuccess({dispatch}, {content, options}: IToast) {
+    const defaultOptions = {}
+    Vue.$toast.success(content, {
+      ...GENERAL_OPTIONS,
+      ...defaultOptions,
```
bulatgab, line 54: Could you double-check, is `defaultOptions` really used throughout the page? It's an empty object.

bulatgab:
The lib looks pretty cool! Also thanks for demoing different options 👏

src/store/notification/index.ts
bulatgab: But this makes the `type` required. Let me think a bit more 🙂

src/store/notification/index.ts
safydy-r: I will remove this

bulatgab (approved):

PR #434, opened 2023-03-17
Feature/CORE-349/remove show hide buttons

safydy-r:
[Changes]
Screen Shot 2023-03-17 at 12 31 42 PM


object-required (approved):
Probably one of the most satisfying kinds of changes, when no code is added 👍

PR #443, opened 2023-04-12
CORE-355 Fix the “Duplicate key detected” message

object-required:
Added the `keyword` of AJV's `ErrorObject` to the element's key because `dataPath` alone may not be unique.


safydy-r (approved):
Nice one @object-required

object-required:
Merged in https://github.com/FLOIP/flow-builder/commit/dd561f167825feac6416435b5d49007842242981

PR #446, opened 2023-04-24
CORE-477 update documentation (round 2)

safydy-r:
I made sure to run the `yarn build-bundle` (just in case I thought), and it came with few changes under `dist/`. That's why we have those updates there.


object-required (approved):
Thank you for the 2nd round 💯 
May we consider adding one more `yarn resolve-build-conflicts` as well?

https://github.com/FLOIP/flow-builder/pull/448

PR #447, opened 2023-04-25
Feature/core-215/core-464/allow multiple container import

lakhbawa:
## Changes
* Removed single flow import validation message

### DIFF
--- a/src/store/flow/views/import.ts
+++ b/src/store/flow/views/import.ts
@@ -249,18 +249,6 @@ export const actions: ActionTree<IImportState, IRootState> = {
         isValid = false
       }
 
-      if (flowContainer.flows?.length > 1) {
-        commit('validation/pushAjvErrorToValidationStatuses', {
-          key,
-          ajvError: {
-            dataPath,
-            keyword,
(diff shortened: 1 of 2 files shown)


bulatgab (approved):

src/store/flow/views/import.ts
```diff
@@ -249,18 +249,6 @@ export const actions: ActionTree<IImportState, IRootState> = {
...
-          key,
-          ajvError: {
-            dataPath,
-            keyword,
-            message: Lang.trans('flow-builder-validation.importer-currently-supports-single-flow-only'),
-          } as ErrorObject,
-        }, {root: true})
-        isValid = false
-      }
-
```
safydy-r, line 263: Thanks for this, but after our pairing, we noticed we should add another changes to avoid the flow[0] in https://github.com/FLOIP/flow-builder/blob/master/src/store/flow/utils/importHelpers.ts#L33

safydy-r (changes requested):
- [ ] Make sure to allow us having multiple flows in the created container

lakhbawa:
declining in favour of https://github.com/FLOIP/flow-builder/pull/451

PR #449, opened 2023-04-25
CORE-477 fix the resource-viewer

safydy-r:
The problem:
If we try to enable the resource viewer on community version by adding `resourceViewer` as part of `enabledFeatures` , we got this error.
Screenshot 2023-04-25 142858

It happens because by default we set `visibleBlocks` as null but we call `visibleBlocks.includes(block.uuid)` in a condition

### DIFF
--- a/src/components/interaction-designer/ResourceViewer.vue
+++ b/src/components/interaction-designer/ResourceViewer.vue
@@ -4,7 +4,7 @@
     <div class="resource-viewer-contents">
       <resource-viewer-block
         v-for="block in activeFlow?.blocks ?? []"
-        v-show="visibleBlocks.includes(block.uuid)"
+        v-show="visibleBlocks === null || visibleBlocks.includes(block.uuid)"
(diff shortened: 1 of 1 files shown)


src/components/interaction-designer/ResourceViewer.vue
```diff
@@ -4,7 +4,7 @@
     <div class="resource-viewer-contents">
       <resource-viewer-block
         v-for="block in activeFlow?.blocks ?? []"
-        v-show="visibleBlocks.includes(block.uuid)"
+        v-show="visibleBlocks === null || visibleBlocks.includes(block.uuid)"
```
safydy-r, line 7: I'm proposing here to always display those block if the visibleBlocks is null, which would allow us to make a test on that edge case scenario.

object-required (approved):

bulatgab (approved):
And we already had the default value for `visibleBlocks` set to `null` 👍

PR #451, opened 2023-04-27
Feature/core 215/core 464/allow multiple container import

lakhbawa:


src/store/flow/views/import.ts
```diff
@@ -249,18 +249,6 @@ export const actions: ActionTree<IImportState, IRootState> = {
         isValid = false
       }
 
-      if (flowContainer.flows?.length > 1) {
```
safydy-r, line 252: let's decline the other PR https://github.com/FLOIP/flow-builder/pull/447 in favor of this one then. 
cc: @bulatgab as you already approved it

src/store/flow/utils/importHelpers.ts
```diff
@@ -27,21 +27,26 @@ export function updateResourcesForLanguageMatch(
...
-    newExistingFlowContainer.flows,
-    (flow) => flow.uuid === newFlowUUID,
-  )
-  if (existingFlowIndex < 0) {
-    newExistingFlowContainer.flows.push(newFlow)
-  } else {
-    newExistingFlowContainer.flows[existingFlowIndex] = newFlow
-  }
-  return newExistingFlowContainer
+  const existingFlowContainerStore = cloneDeep(existingFlowContainer);
```
safydy-r, line 30: why did we change `newExistingFlowContainer` into `existingFlowContainerStore`? 
- let's avoid using `store` prefix as this brings confusion to vuex stores, we have a lot of that in this repo
- unless you have a strong opinion on the renaming, I'd recommend we keep `newExistingFlowContainer` or use `clonedExistingFlowContainer`

- [ ] improve the variable name

src/store/flow/utils/importHelpers.ts
```diff
@@ -27,21 +27,26 @@ export function updateResourcesForLanguageMatch(
...
+
+  const newFlows = get(newFlowContainer, 'flows', []);
```
safydy-r, line 32: We don't put `;` for JS codes

- [ ] make the end of line uniform, remove `;`

src/store/flow/utils/importHelpers.ts
```diff
@@ -27,21 +27,26 @@ export function updateResourcesForLanguageMatch(
...
+
+  newFlows.forEach((newFlow) => {
+    const newFlowUUID = get(newFlow, 'uuid');
```
bulatgab, line 35: we probably don't need `get` because a flow always has a uuid.
maybe use `newFlow.uuid` + inline it?
(and the same for `newFlows` above)

lakhbawa:
ready for second review @safydy-r @bulatgab

src/store/flow/flow.ts
```diff
@@ -178,7 +178,7 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
    * Persistence for IMPORT action
    */
   async flow_persistImport({getters, commit, dispatch}, {persistRoute, flowContainer}): Promise<IContext | null> {
-    flowContainer = mergeFlowContainer(cloneDeep(getters.activeFlowContainer), flowContainer)
+    flowContainer = mergeFlowContainer(getters.activeFlowContainer, flowContainer)
```
safydy-r, line 181: 👍

safydy-r (approved):
thanks for address pr review comments ;)

bulatgab (approved):
👍

PR #453, opened 2023-04-27
Feature/core 215/core 483/open main flow after container import

lakhbawa:
* Create a method to create flow stack from container
*


src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+  uuid: string;
+  blocks: CreateFlowStackBlock[];
+}
+
+interface CreateFlowStackData {
+  flows: CreateFlowStackFlow[];
+}
+
+export function createContainerFlowStack(json_data: any, flow_stack: string[]): void {
+  const flows = json_data.flows;
```
safydy-r, line 70: We don't put `;` for JS codes

- [ ] make the end of line uniform, remove `;`

src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+  const visitedFlows: string[] = [];
```
safydy-r, line 71: if this is a list of flow uuid, let's improve this as follow
```
const visitedFlowUuids: IFlow['uuid'][] = []
```

src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+
+  function process_flow(flow: CreateFlowStackFlow): void {
```
safydy-r, line 73: in general, almost all functions we write are about processing something. This naming is not very explanatory.

- [ ] improve the naming to give more context about what it does

safydy-r: Don't hesitate to leave a JS_doc if you think we need more longer context to help devs

src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+    visitedFlows.push(flow.uuid);
+    // eslint-disable-next-line no-restricted-syntax
```
safydy-r, line 75: could you remind me what error do we got if we remove those `// eslint-disable-next-line no-restricted-syntax` ? I'm wondering if there is a better way to handle that

src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+    for (const block of flow.blocks) {
+      if (block.type === 'Core.RunFlow') {
```
safydy-r, line 77: let's use the `BLOCK_TYPE` from run flow store
```
import {BLOCK_TYPE as RUN_FLOW_BLOCK_TYPE } from '@/store/flow/block-types/Core_RunFlowBlockStore'

...
if(type === RUN_FLOW_BLOCK_TYPE)
```

src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+interface CreateFlowStackFlow {
+  uuid: string;
+  blocks: CreateFlowStackBlock[];
+}
+
+interface CreateFlowStackData {
+  flows: CreateFlowStackFlow[];
+}
+
+export function createContainerFlowStack(json_data: any, flow_stack: string[]): void {
```
safydy-r, line 69: This method is quite complex to understand. I'd recommend we create an unit testing to cover this, in case we make some changes later, the test cases will confirm we get it right

- [ ] add unit testing for this

src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+      process_flow(flow);
+    }
+  }
+}
+
+export function getLastItemFromContainerFlowStack(flow_stack: string[]): string {
+  return flow_stack[flow_stack.length - 1];
+}
+export function createContainerFlowStackAndReturnLastItem(json_data: any): string {
+    const flow_stack: string[] = [];
```
bulatgab, line 104: AFAIK, our convention is to use camelCase for variable names 😉
(one of the exceptions is the data sent from/to a server)

bulatgab: the same for `process_flow`, etc.

src/store/flow/utils/importHelpers.ts
bulatgab: @lakhbawa could you also add a JSDoc to this function explaning how it works and behaves?

src/store/flow/utils/importHelpers.ts
```diff
@@ -49,6 +49,64 @@ export function mergeFlowContainer(
...
+    if (isEmpty(json_data)) { return ''; }
+    createContainerFlowStack(json_data, flow_stack);
+    return getLastItemFromContainerFlowStack(flow_stack);
+}
```
bulatgab, line 108: I would also recommend using a more functional-programming style, like:
```
export function createContainerFlowStackAndReturnLastItem(json_data: any): string {
    if (isEmpty(json_data)) {
        return ''
    }
    const flowStack = createContainerFlowStack(json_data)
    return flowStack.pop()!
}
```
Here the `createContainerFlowStack` is taking an input and returning a value, rather than mutating the input.
Also, as the `flowStack` is not needed for anything else, we can `pop` the last item - that's simpler.

bulatgab: That would mean refactoring the `createContainerFlowStack`

src/store/flow/utils/importHelpers.ts
bulatgab: In my understanding, you're creating a hierarchy of flows - how they use each other via RunFlowBlock. But the details of this hierarchy are unclear.

What if a flow has multiple RunFlowBlocks, and they have their own children, etc.
How would they be ordered in the stack?
Will all of the flows in the container be included in the stack?
Or only one 'lineage' of flows from top to botton, ignoring some siblings?

bulatgab (changes requested):
Please see code comments

src/store/flow/utils/importHelpers.ts
safydy-r: fyi: we have test sample here https://github.com/FLOIP/flow-builder/blob/master/tests/unit/components/interaction-designer/block-editors/choices/expressionTransformers.spec.ts

lakhbawa:
ready for second review @safydy-r @bulatgab

src/store/flow/utils/importHelpers.ts
lakhbawa: Thanks @bulatgab @safydy-r added description + test

src/store/flow/utils/importHelpers.ts
lakhbawa: fixed it by using array iterators

src/store/flow/utils/importHelpers.ts
lakhbawa: Thanks bulat, i was looking for the naming convention

src/store/flow/utils/importHelpers.ts
lakhbawa: Thanks, Bulat, refactored createContainerFlowStack to return

tests/unit/flow-import/flowImport.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      }
+    `
+    const jsonDataParsed = JSON.parse(jsonData)
+
+    const contextObject = {
+    name: 'TODO',
+    description: 'TODO',
+    vendor_metadata: {},
+    flows: jsonDataParsed.flows,
+  } as unknown as IContext
```
lakhbawa, line 59: not fully sure, how it works, but took some inspiration from existing code

tests/unit/flow-import/flowImport.spec.ts
```diff
@@ -0,0 +1,63 @@
+import {createContainerFlowStack} from '@/store/flow/utils/importHelpers'
+import {IContext} from '@floip/flow-runner'
+
+describe('create_flow_stack', () => {
+  test('should create a flow stack with correct order', () => {
```
safydy-r, line 5: thanks for providing this test 🥇

tests/unit/flow-import/flowImport.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const contextObject = {
+    name: 'TODO',
+    description: 'TODO',
+    vendor_metadata: {},
+    flows: jsonDataParsed.flows,
+  } as unknown as IContext
+    const flowStack = createContainerFlowStack(contextObject)
+    expect(flowStack).toEqual(['flow3', 'flow2', 'flow4', 'flow1', 'flow5'])
+  })
```
safydy-r, line 62: @lakhbawa , as I mentioned this `createContainerFlowStack` is a bit complex, I'd suggest we add another test case to cover "multiple flows importation where there is no dependencies between flows" (I think, now the method would work, but later when someone is updating this method, we gonna need that test case to confirm)

- [ ] add another test case

safydy-r (changes requested):
lgtm, thanks. Just have 01 request to cover the unit testing

tests/unit/flow-import/flowImport.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  test('should create a flow stack with correct order_when_there_are_sibling_flows', () => {
+    const jsonData = `
+      {
+        "flows": [
```
safydy-r, line 67: 👍

safydy-r (approved):
lgtm now

PR #454, opened 2023-05-01
Feature/CORE-461/fix long expression dropdown

safydy-r:
Overview
core-461

**Note**: For now, when we have long expression, the dropdown seems to be aligned on the right. I tried to find the issue there, but couldn't identify it. So, I'd suggest we keep it as it is for now. After all, everything is visible to the user for now.


src/components/common/ExpressionInput.vue
```diff
@@ -108,8 +110,9 @@ export class ExpressionInput extends mixins(Lang) {
    */
   portAutoSuggestContent(): void {
     // override position which came from AutoSuggest original code
+    this.autoSuggestDropdown.style.position = 'relative'
```
safydy-r, line 113: by default, the dropdown has "absolute" position, and some values on left & top.
The solution is just to get rid of that to have relative position, and then allow flex display to handle a column display https://getbootstrap.com/docs/4.0/utilities/flex/#direction

lakhbawa (approved):
looks good to me.

object-required (approved):

PR #455, opened 2023-05-01
Feature/core 215/multiple flow import  integration

lakhbawa:


src/store/flow/utils/importHelpers.ts
```diff
@@ -27,21 +28,69 @@ export function updateResourcesForLanguageMatch(
...
+  const flowStack: IFlow['uuid'][] = []
+
+  /*
+  * The recursive function processFlowForCreatingFlowStack takes a flow object as an argument and starts iterating over its blocks. If a block has a type of "Core.RunFlow,"
+  * it means that this block is calling another flow,
+  * so the function recursively calls itself with the child flow as an argument.
+  * This ensures that all the nested flows are processed before moving back to the parent flow.
+  *
+  * Once all the nested flows are processed, the parent flow is added to the flowStack array.
+  * */
```
bulatgab, line 63: Thank you, Lakh. This JSDoc + unit tests help understand the algorithm 👍

(I guess we could say that we are looking for _the last flow in the list that doesn't have any parent_, right?)

bulatgab (approved):

src/store/flow/utils/importHelpers.ts
lakhbawa: Yes, that's correct @bulatgab

safydy-r (approved):

PR #459, opened 2023-05-16
Feature/core 538/core 539/updated undo redo prototype

bulatgab:
Demo

https://github.com/FLOIP/flow-builder/assets/18130716/8ea5f059-aee0-4e26-9190-d09fd05fa47f


src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -314,12 +320,14 @@
...
-    <div v-if="isBuilderCanvasEnabled" class="tree-builder-toolbar-alerts w-100">
+    <div
+      v-if="isBuilderCanvasEnabled"
+      class="tree-builder-toolbar-alerts w-100">
       <selection-banner
         v-if="isEditable"
         @updated="handleHeightChangeFromDOM" />
       <error-notifications @updated="handleHeightChangeFromDOM" />
-      <block-editor v-if="isBlockEditorOpen"/>
+      <block-editor v-if="isBlockEditorOpen" />
```
bulatgab, line 330: _ESLint fixes_

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -314,12 +320,14 @@
         </div>
       </div>
     </div>
-    <div v-if="isBuilderCanvasEnabled" class="tree-builder-toolbar-alerts w-100">
+    <div
+      v-if="isBuilderCanvasEnabled"
+      class="tree-builder-toolbar-alerts w-100">
```
bulatgab, line 325: _ESLint fixes_

src/components/interaction-designer/block/BlockToolbar.vue
```diff
@@ -84,40 +85,32 @@ export class BlockToolbar extends mixins(Lang) {
...
+    await this.takeSnapshot()
     this.$emit('after-delete')
   }
 
-  handleDuplicateBlock(): void {
-    this.flow_duplicateBlock({blockId: this.block.uuid}).then((duplicatedBlock) => {
-      this.$router.replace({
-        name: 'block-selected-details',
-        params: {blockId: duplicatedBlock.uuid},
-      })
```
bulatgab, line 99: used the chance to refactor `.then()` into `await`

src/store/flow/flow.ts
```diff
@@ -572,20 +575,21 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
-    forEach(state.selectedBlocks, (blockId: IBlock['uuid']) => {
-      dispatch('flow_removeBlock', {blockId})
-    })
+    const removeBlock_actions = state.selectedBlocks.map((blockId: IBlock['uuid']) =>
+      dispatch('flow_removeBlock', {blockId}))
+
+    await Promise.all(removeBlock_actions)
 
     state.selectedBlocks = []
   },
```
bulatgab, line 584: This kind of `Promise.all` logic (here and elsewhere) is needed to make sure that everything is done before we save the snapshot

src/store/flow/flow.ts
```diff
@@ -572,20 +575,21 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
...
-      const duplicatedBlock: IBlock = await dispatch('flow_duplicateBlock', {blockId})
-      newBlocksUuid.push(duplicatedBlock.uuid)
-    })
-    state.selectedBlocks = newBlocksUuid
+    const duplicateBlock_actions = state.selectedBlocks.map((blockId: IBlock['uuid']) =>
+      dispatch('flow_duplicateBlock', {blockId}))
+
+    const duplicatedBlocks: IBlock[] = await Promise.all(duplicateBlock_actions)
+
+    state.selectedBlocks = duplicatedBlocks.map(block => block.uuid)
```
bulatgab, line 592: `await Promise.all` logic

src/components/interaction-designer/block/BlockToolbar.vue
safydy-r: @bulatgab , I'm curious why do we need to replace then to await?
- with `then`, we only call the instruction if the main promise was a success
- with await, we always call the next instruction even the main promise was a failure (thus, duplicatedBlock might not be defined)

src/assets/messages.json
```diff
@@ -1282,7 +1282,7 @@
     "selected": "selected",
     "selected-groups": "Selected Groups",
     "selected-subscribers": "Selected contacts",
-    "selected-x-blocks": "You have seletected :x block(s)",
+    "selected-x-blocks": "You have selected :x block(s)",
```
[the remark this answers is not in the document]
bulatgab: Not at the moment, but now it's updated. Thanks!

src/components/interaction-designer/toolbar/UndoRedoButtonGroup.vue
```diff
@@ -0,0 +1,39 @@
...
+<script setup lang="ts">
+import {computed} from 'vue'
+import {useStore} from '@/store/useStore'
+
+const store = useStore()
+
+const canUndo = computed<boolean>(() => store.getters['undoRedo/canUndo'])
+const canRedo = computed<boolean>(() => store.getters['undoRedo/canRedo'])
+
+// const undoTooltip = computed<string | undefined>(() => store.getters['undoRedo/undoTooltip'])
```
safydy-r: I think designer has agreed on texts here? We can just use localized texts

- [ ] enable the tooltip

src/font-awesome-icon.ts
```diff
@@ -14,6 +14,8 @@ import {
   faAngleDown,
   faAngleUp,
   faTimesCircle as faSolidTimesCircle,
+  faRedo,
```
safydy-r, line 17: 👍

safydy-r (approved):
no big concern to me, pre-approving. I have a question about replacing `then` into `await`
https://github.com/FLOIP/flow-builder/pull/459#discussion_r1197017874

src/components/interaction-designer/block/BlockToolbar.vue
bulatgab: To me it was about code style only. I think we are mainly using `async+await` in our codebase rather than `.then()`.

In terms of equivalence - I might be wrong but I think if the promise rejects on this line:
```
const duplicatedBlock = await this.flow_duplicateBlock({blockId: this.block.uuid})
```
it would throw an error and the rest of the code in the function would be skipped.

I guess we could add try-catch indeed (as it's often done around await) just to be on the safe side! Would that work?

src/store/undoRedo/index.ts
```diff
@@ -0,0 +1,143 @@
...
+      console.warn('Cannot undo, we have already reached the beginning of the history')
+    }
+  },
+  redo(state) {
+    if (state.position < state.snapshots.length - 1) {
+      state.position += 1
+    } else {
+      console.warn('Cannot redo, we have already reached the end of the history')
+    }
+  },
```
bulatgab, line 56: I kept the 'undo'/'redo' naming for these mutations, but to be more accurate, we might consider something like 'decrementPosition'/'incrementPosition' or 'decPosition'/'incPosition', although I don't like that very much :)

src/components/interaction-designer/toolbar/UndoRedoButtonGroup.vue
bulatgab: Sorry! I forgot to push the latest change. Updated now. Just 2 little commits:
https://github.com/FLOIP/flow-builder/pull/459/commits/5b118e9242a0d7caa1a8e027ebbf72e7045d1f2a
https://github.com/FLOIP/flow-builder/pull/459/commits/2da74b8367043237c16259fafa27ed5ab65eb9de

src/components/interaction-designer/block/BlockToolbar.vue
bulatgab: Fixed

PR #460, opened 2023-05-19
CORE-538, CORE-567 State management plugin (spike)

object-required:
- [x] `getChangedKeys` does not work well with arrays and nested objects
- [ ] subsequent changes are considered one, e.g., the user edits text in several seconds and wants to undo the change
- [x] fix block deletion
- [x] store the history of `previouslyChangedKeys` as well


src/main.ts
```diff
@@ -41,7 +42,9 @@ Vue.use(Toast, VUE_TOAST_GENERAL_OPTIONS)
 Vue.config.productionTip = false
 
 // todo CORE-155: inline 'store' after migration to Vue3
-const store = new Vuex.Store({})
+const store = new Vuex.Store({
+  plugins: [VuexUndoRedoPlugin],
```
safydy-r, line 46: @object-required , If I'm not wrong this would require we do similar thing on voto5?
if YES, Is there a better way to just only do it within the community version, and not replicate the declaration on the consumer side?

object-required: Great point, I moved plugin registration to `TreeBuilderToolbar.created()` lifecycle hook --
https://github.com/FLOIP/flow-builder/blob/d3fcefec4ca8e1bcb727aefd9753eb0c89282c89/src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue#L372-L374

src/components/interaction-designer/toolbar/TreeBuilderToolbar.vue
```diff
@@ -368,6 +370,17 @@ export class TreeBuilderToolbar extends mixins(Routes, Permissions, Lang) {
...
+    const $store = this.$store
+    const moduleName = 'undoRedo'
+
+    if (!$store.hasModule(moduleName)) {
+      $store.registerModule(moduleName, undoRedoModule)
+    }
+
+    VuexUndoRedoPlugin(this.$store)
+  }
+
```
bulatgab, line 383: Would it be a good idea to extract this logic into a separate function inside `vuex-undo-redo-plugin.ts` file, because `TreeBuilderToolbar.vue` is pretty big?

bulatgab: Wait, this is also registering the module that is in a separate file. Never mind

src/store/undoRedo/getChangedKeys.ts
```diff
@@ -0,0 +1,18 @@
...
+
+      if (typeof aVal === 'object' && typeof bVal === 'object') {
+        changedKeys.push(...getChangedKeys(aVal as object, bVal as object).map((k) => `${key}.${k}`))
+      } else if (aVal !== bVal) {
+        changedKeys.push(key)
+      }
+    })
+
+  return changedKeys
+}
```
bulatgab, line 18: Should we use stricter types for the args so that `getChangedKeys` is not called with primitives?
I've played with this a bit and also noticed that `=== object` passes through `null`.
Suggestion:
```
// eslint-disable-next-line import/prefer-default-export
export function getChangedKeys(a: object, b: object): string[] {
  const changedKeys: string[] = []

  Object.keys({...a, ...b})
    .forEach((key) => {
      const aVal = (a as Record<string, unknown>)[key]
      const bVal = (b as Record<string, unknown>)[key]

      if (typeof aVal === 'object' && typeof bVal === 'object' && aVal !== null && bVal !== null) {
        changedKeys.push(...getChangedKeys(aVal, bVal).map((k) => `${key}.${k}`))
      } else if (aVal !== bVal) {
        changedKeys.push(key)
      }
    })

  return changedKeys
}
```
Luckily, the null doesn't currently break things as spreading it in `Object.keys({...null, ...null})` gives an empty array.
So the suggested code doesn't change much (just marginally better typing?)

src/store/undoRedo/index.ts
```diff
@@ -54,81 +101,71 @@ export const mutations: MutationTree<IUndoRedoState> = {
...
 
   /**
-   * Clear undos and redos, keep current state
+   * Reset the history of changes to the current state
    */
-  async clearAllHistory({commit, dispatch}) {
-    const snapshot = await dispatch('cloneState')
-    commit('resetSnapshots', snapshot)
+  async resetHistory({commit, rootState}) {
+    setImmediate(() => {
```
bulatgab, line 164: could you plase add a comment about why we need setImmediate

src/store/undoRedo/index.ts
```diff
@@ -29,17 +54,39 @@ export const getters: GetterTree<IUndoRedoState, IRootState> = {
...
   },
+  currentSnapshot(state): ISnapshot {
+    return state.snapshots[state.position]
+  },
+  previousSnapshot(state): ISnapshot | null {
+    return state.snapshots[state.position - 1] ?? null
+  },
+  futureSnapshot(state): ISnapshot | null {
+    return state.snapshots[state.position + 1] ?? null
+  },
```
bulatgab, line 65: 👍
I guess we can rewrite `canUndo` in the future, like `return this.previousSnapshot !== null`
And the same for `canRedo`

bulatgab (approved):

PR #461, opened 2023-05-23
Feature/core 538/core 541/undoable simple block config changes

bulatgab:


src/components/interaction-designer/block-editors/GroupMembershipEditor.vue
```diff
@@ -175,19 +180,20 @@ export class GroupMembershipEditor extends mixins(Lang) {
...
   }
 
-  onGroupAdd(name: string): void {
+  async onGroupAdd(name: string): Promise<void> {
     const newGroup: IGroupMembership = {
       group_key: name,
       group_name: name,
     }
 
     this.userAddedGroups.push(newGroup)
```
bulatgab, line 189: I have kept `userAddedGroups` local which makes adding groups undoable. Hope that is okay.

src/components/interaction-designer/block-editors/GroupMembershipEditor.vue
```diff
@@ -28,14 +29,15 @@
...
             :class="{invalid: isValid === false}"
             :is-loading="hasGroupsLoading"
             :options="groupOptions"
             :taggable="!availableGroups"
             :multiple="true"
             track-by="group_key"
             label="group_name"
+            :value="selectedGroups"
+            @input="setSelectedGroups"
             @seach-change="onSearchChange"
```
bulatgab, line 41: @object-required there is a typo in `@search-change` which probably means this is not used anywhere. Should we delete the related downstream logic? Or keep it in case the lib's users need this functionality?

PR #462, opened 2023-05-26
CORE-538, CORE-547 Reset change history on save

object-required:
The `undoRedo/resetHistory` action is now dispatched from `flow/persistFlowAndHandleUiState` instead of `TreeBuilderToolbar` component.

### DIFF
--- a/src/store/builder/index.ts
+++ b/src/store/builder/index.ts
@@ -302,6 +302,8 @@ export const actions: ActionTree<IBuilderState, IRootState> = {
         timeout: 3000,
         hideProgressBar: true,
       })
+
+      dispatch('undoRedo/resetHistory', null, {root: true})
     }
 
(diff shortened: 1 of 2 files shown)


bulatgab (approved):

src/store/builder/index.ts
```diff
@@ -302,6 +302,8 @@ export const actions: ActionTree<IBuilderState, IRootState> = {
         timeout: 3000,
         hideProgressBar: true,
       })
+
+      dispatch('undoRedo/resetHistory', null, {root: true})
```
safydy-r, line 306: @object-required , could you pls confirm this is working when we do have auto save:
- from the resource viewer page (after we change a resource & being inactive for a moment)
- from the tree build Voto5 (After being inactive for a moment)

PR #464, opened 2023-05-29
CORE-538, CORE-545 Block connections

object-required:


safydy-r (approved):
happy this worked & was easier than we thought :D

bulatgab:
@object-required 
I don't understand how it works 😅 Could you please explain?

bulatgab:
I've noticed that if we remove the connection by clicking on the exit, it can't be undone. Could you please check?
Screenshot 2023-05-30 at 1 14 34 PM

bulatgab (changes requested):
- [ ] please leave a comment on how/why deactivateAllConnections removes the connection
- [ ] fix the 'undo remove connection'

src/store/undoRedo/index.ts
```diff
@@ -38,6 +38,12 @@ function getChangedModulesKeys(snapshotA: ISnapshot, snapshotB: ISnapshot): stri
   return getChangedKeys(modulesA, modulesB)
 }
 
+function shouldAlwaysTriggerSnapshot(changedKeys: string[]): boolean {
+  return changedKeys.some((key) => (
+    /^flows.flows.\d+.blocks.\d+.exits.\d+.destination_block$/.test(key)
+  ))
+}
+
```
object-required, line 46: @bulatgab creating and deleting block connections right away fell under the "we changed the same thing" category; I created a list of overrides with store paths that should always produce a new snapshot.

PR #465, opened 2023-05-29
CORE-538, CORE-544 Handle move block action

object-required:
[Screencast from 2023-05-29 04:24:40 PM.webm](https://github.com/FLOIP/flow-builder/assets/3971957/0123ea66-3d7d-4874-89d8-ab88c0b71486)

### DIFF
--- a/src/components/common/PlainDraggable.vue
+++ b/src/components/common/PlainDraggable.vue
@@ -35,11 +35,29 @@ export class PlainDraggable extends mixins(Lang) {
     console.debug('PlainDraggable.vue', 'mounted')
 
     // this.$nextTick(() => {
+
+    this.draggable = new PlainDraggableLib(this.$el, this.plainDraggableLibOptions)
+    // draggable.rect.{left,top,x,y,...}
+
+    // todo: why this doesn't work?
+
+    // this.draggable.snap = {x: 50, y: 50, width: 50, height: 50}
+    // this.draggable.snap = {step: 21}
+
+    this.handleInitialized()
+    // })
+  }
+
+  updated(): void {
+    this.draggable.setOptions(this.plainDraggableLibOptions)
+  }
+
+  get plainDraggableLibOptions(): Record<string, unknown> {
     const handle = this.dragHandleId !== undefined
       ? document.getElementById(this.dragHandleId)
       : this.$el.querySelectorAll('.draggable-handle')[0]
 
-    this.draggable = new PlainDraggableLib(this.$el, {
+    return {
       containment: document.querySelector('.builder-canvas'),
       autoScroll: true,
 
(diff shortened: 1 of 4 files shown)


src/components/common/PlainDraggable.vue
```diff
@@ -35,11 +35,29 @@ export class PlainDraggable extends mixins(Lang) {
...
+    // this.draggable.snap = {x: 50, y: 50, width: 50, height: 50}
+    // this.draggable.snap = {step: 21}
+
+    this.handleInitialized()
+    // })
+  }
+
+  updated(): void {
+    this.draggable.setOptions(this.plainDraggableLibOptions)
+  }
```
object-required, line 53: `<Block>` passes updated coordinates to `<PlainDraggable startX startY>` which allows us to update block position when needed.

src/store/builder/index.ts
```diff
@@ -309,6 +304,14 @@ export const actions: ActionTree<IBuilderState, IRootState> = {
...
     return newFlowContainer ?? undefined
   },
+
+  setBlockPositionTo({commit, rootGetters}, {position: {x, y}, block}) {
+    commit('flow/flow_updateBlockCoordinates', {
+      flowId: rootGetters['flow/activeFlow'].uuid,
+      blockId: block.uuid,
+      coordinates: {x, y},
+    }, {root: true})
+  },
```
object-required, line 314: Making this an action allows committing a mutation in the neighbor `flow` module.

src/store/builder/index.ts
```diff
@@ -140,11 +140,6 @@ export const mutations: MutationTree<IBuilderState> = {
     operations[operation.kind] = operation
   },
 
-  setBlockPositionTo(state, {position: {x, y}, block}) {
-    block.ui_metadata.canvas_coordinates.x = x
-    block.ui_metadata.canvas_coordinates.y = y
-  },
```
object-required, line 146: Moved to `flow/flow_updateBlockCoordinates` to trigger a mutation that `undoRedo` can handle.

src/store/flow/flow.ts
```diff
@@ -202,6 +202,21 @@ export const mutations: MutationTree<IFlowsState> = {
...
+    const flow: IFlow = findFlowWith(flowId, state as unknown as IContext)
+    const block: IBlock = findBlockWith(blockId, flow)
+
+    if (!block.ui_metadata) {
+      Vue.set(block, 'ui_metadata', {
+        canvas_coordinates: coordinates,
+      })
+    } else {
+      Vue.set(block.ui_metadata, 'canvas_coordinates', coordinates)
+    }
```
bulatgab, line 218: A theoretical question @object-required @safydy-r 
What would happen if we do:
```
Vue.set(block, 'ui_metadata', {canvas_coordinates: coordinates})
```
without if-else? I mean what would happen in terms of Vue reactivity if `ui_metadata.canvas_coordinates` exists? What's the danger of vue-setting a higher level object?

I think it's possible it React, for example, right? Even preferrable, I think.
Do you know how Vue handles this? I couldn't find any article on this (maybe I was googling it wrong) - could you help, please?

bulatgab (approved):
Clean refactoring/fix, thank you.

Left an off-topic question.

src/store/builder/index.ts
safydy-r: Nice solution, I agree

safydy-r (approved):
Joining @bulatgab 's comment, this is a clean one. Thanks @object-required 👍

PR #467, opened 2023-06-01
Feature/CORE-538/CORE-575/prepare e2e testing dependencies

safydy-r:
[overview of the 1st test]
core-575


object-required (approved):
This looks amazing :fire:

cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
safydy-r, line 1: auto generated by cypress as examples

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "esnext",
+    "target": "ES2018",
```
safydy-r, line 3: this is important to allow cypress understanding the `??` & `?.` operators

src/views/NewFlow.vue
```diff
@@ -26,7 +26,7 @@
 
               <a
                 :href="route('flows.editFlow', {flowId: activeFlow.uuid, component: 'builder', mode: 'edit'})"
-                class="btn btn-primary"
+                class="btn btn-primary cy--create--btn"
```
bulatgab, line 29: what do you think about using `data-` attributes in the future and maybe even custom commands like `getBySel`? (see here: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements)

src/main.ts
```diff
@@ -51,6 +61,14 @@ async function main(): Promise<void> {
     provide: {store},
     render: (h) => h(Vue.extend(App)),
   }).$mount('#app')
+
+  if (window?.Cypress !== undefined) {
+    console.debug('Cypress detected, exposing app on window')
+    // This will allow us to test vuex store in Cypress
+    window.store = store
```
bulatgab, line 68: I'm not a fan of global objects but I couldn't find a better approach 👍

bulatgab (approved):
So cool to see e2e tests 👍
Thanks!

src/main.ts
safydy-r: Same here, I look for alternative but couldn't find a working one. 
That's why I put it under the condition `window?.Cypress !== undefined` to not set that global var on prod

src/views/NewFlow.vue
safydy-r: Good point, I forgot to check to recommendation. I think I'm gonna make changes at this PR. Thanks @bulatgab

cypress/e2e/undo-redo/create-new-flow.cy.js
```diff
@@ -0,0 +1,38 @@
...
+    const getStore = () => cy.window().its('store')
+    const flowModuleState = () => getStore().its('state.flow')
+    const flowsListState = () => flowModuleState().its('flows')
+    flowsListState().should('have.length', 0)
+    cy.get('[data-cy="create-flow--btn"]').click()
+
+    // ############ Create flow page ############
+    const flowName = 'test flow'
+    const interactionTimeout = 1000
+    cy.get('[data-cy="create--btn"]').click()
```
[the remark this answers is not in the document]
bulatgab: Thanks :)

PR #468, opened 2023-06-02
Feature/core 538/core 580/improve ux on save action

safydy-r:
[overview]
To note:
- we make sure we don't capture another snapshot after SAVE action, we should patch the previous one instead
- we disable the EDIT during the SAVE action to avoid weird side effect

core-580


src/store/builder/index.ts
```diff
@@ -297,11 +298,10 @@ export const actions: ActionTree<IBuilderState, IRootState> = {
         timeout: 3000,
         hideProgressBar: true,
       })
-
-      dispatch('undoRedo/resetHistory', null, {root: true})
```
safydy-r, line 301: do not clear the history as discussed with PM

src/store/undoRedo/index.ts
```diff
@@ -115,15 +124,29 @@ export const actions: ActionTree<IUndoRedoState, IRootState> = {
       flows: rootState.flow,
     })
```
bulatgab, line 125: would `newSnapshot` be a better name instead of `nextSnapshot`?
'next' is associated with the 'next in the stack' for me, whereas 'new' - is newly created, not placed anywhere yet

src/store/undoRedo/index.ts
```diff
@@ -115,15 +124,29 @@ export const actions: ActionTree<IUndoRedoState, IRootState> = {
...
 
+    let isLastFlowStateFromPersistence = !getters.hasCurrentSnapshot
+    const currentSavedAt = getters.currentSnapshot?.modules.flows?.savedAt
+    const nextSavedAt = nextSnapshot?.modules.flows?.savedAt
+    if (currentSavedAt !== undefined && nextSavedAt !== undefined && currentSavedAt !== nextSavedAt) {
+      isLastFlowStateFromPersistence = true
+    }
```
bulatgab, line 132: would this be equivalent? 🤔
```
const isLastFlowStateFromPersistence  = nextSavedAt !== currentSavedAt
```
`nextSavedAt` can't be different AND undefined - because we are never clearing it

src/store/undoRedo/index.ts
```diff
@@ -115,15 +124,29 @@ export const actions: ActionTree<IUndoRedoState, IRootState> = {
       flows: rootState.flow,
     })
 
+    let isLastFlowStateFromPersistence = !getters.hasCurrentSnapshot
+    const currentSavedAt = getters.currentSnapshot?.modules.flows?.savedAt
+    const nextSavedAt = nextSnapshot?.modules.flows?.savedAt
```
bulatgab, line 129: would this be better naming?
`isLastFlowStateFromPersistence` -> `isCurrentFlowStateFromPersistence` ('last' is associated with the snapshot stack to me again, but we are talking about state)
`currentSavedAt` -> `oldSavedAt`
`nextSavedAt` -> `newSavedAt`

src/store/undoRedo/index.ts
```diff
@@ -115,15 +124,29 @@ export const actions: ActionTree<IUndoRedoState, IRootState> = {
       flows: rootState.flow,
     })
 
+    let isLastFlowStateFromPersistence = !getters.hasCurrentSnapshot
```
bulatgab, line 127: this line is unclear 🤔
Could you please clarify?

src/store/undoRedo/index.ts
```diff
@@ -115,15 +124,29 @@ export const actions: ActionTree<IUndoRedoState, IRootState> = {
...
+    const nextSavedAt = nextSnapshot?.modules.flows?.savedAt
+    if (currentSavedAt !== undefined && nextSavedAt !== undefined && currentSavedAt !== nextSavedAt) {
+      isLastFlowStateFromPersistence = true
+    }
+
+    // ####### force patch ##############
+    if (isLastFlowStateFromPersistence) {
+      commit('patchSnapshot', nextSnapshot)
+      return
+    }
```
bulatgab, line 138: can we merge this with the `commit` below?

bulatgab (changes requested):
Thank you @safydy-r for sharing the PR.
Could you please consider the proposed changes?
(Or refactor it some other way - because it was a bit hard to me to understand the current code)

src/store/undoRedo/index.ts
```diff
@@ -67,6 +67,15 @@ export const getters: GetterTree<IUndoRedoState, IRootState> = {
...
   },
+  hasCurrentSnapshot(_state, getters): boolean {
+    return Boolean(getters.currentSnapshot)
+  },
+    hasPreviousSnapshot(_state, getters): boolean {
+    return Boolean(getters.previousSnapshot)
+  },
+  hasFutureSnapshot(_state, getters): boolean {
+    return Boolean(getters.futureSnapshot)
+  },
```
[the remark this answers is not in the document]
object-required: The point of `can*` was providing boolean types, but I guess it's ok to inline those because they are currently used once each.

src/store/undoRedo/index.ts
object-required: I agree, `newSnapshot` would read better.

src/store/undoRedo/index.ts
safydy-r: yeah, I agree. 
At the beginning, we don't have currentSnapshot yet, if that's the case, we consider the last flow state was from a persistence action, (eg: after we load a flow)
The original code was
```
let isLastFlowStateFromPersistence = false
if(!getters.hasCurrentSnapshot) {
    isLastFlowStateFromPersistence = true
}
````
Without this, we will have 02 snapshots after the flow load instead of 01.

I think I can leave a comment here to clarify it more

- [ ] leave comment

File: src/store/undoRedo/index.ts (hunk 3 above)
Comment (safydy-r) [reply src/store/undoRedo/index.ts:132]: let me reorder the condition if that's more easier to read, no major concern here ;)
But I'd like to keep the `nextSavedAt !== undefined` for precaution

File: src/store/undoRedo/index.ts (hunk 6 above)
Comment (safydy-r) [reply src/store/undoRedo/index.ts:138]: I'd prefer to leave the "early return" pattern, and maybe update the existing code to something like
```
if(getters.hasFutureSnapshot) {
      commit('addSnapshot', nextSnapshot)
      return
}

const hasSpecialKeys = ...
if(changedKeys) {
      commit('addSnapshot', nextSnapshot)
      return
}

...
commit('patchSnapshot', nextSnapshot)
````

That way we don't need to way for the whole `hasDifferentKeys || getters.hasFutureSnapshot || hasSpecialKeys` to be ready before we decide
Your thoughts @object-required , @bulatgab

src/store/undoRedo/index.ts
safydy-r: renamed !

src/store/undoRedo/index.ts
safydy-r: re-ordered!

src/store/undoRedo/index.ts
bulatgab: To me - that would be another option, yes. Because then we would be doing a _consistent_ early return 👍

src/store/undoRedo/index.ts
bulatgab: The new code looks more understandable to me, thank you Safidy

bulatgab (approved):
Thanks a lot for the changes; approving :)

The last things that can be looked at are the early return thing and the `can*` getters duplication - if you have time.

PR #471, opened 2023-06-05
Feature/CORE-538/CORE-579/adjust undo redo on resource view page

safydy-r:
[screen capture]
core-579

### DIFF
--- a/src/components/interaction-designer/ResourceViewer.vue
+++ b/src/components/interaction-designer/ResourceViewer.vue
@@ -4,7 +4,7 @@
     <div class="resource-viewer-contents">
       <resource-viewer-block
         v-for="block in activeFlow?.blocks ?? []"
-        v-show="visibleBlocks.includes(block.uuid)"
+        v-show="visibleBlocks === null || visibleBlocks.includes(block.uuid)"
         :id="`block/${block.uuid}`"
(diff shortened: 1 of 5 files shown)


src/components/interaction-designer/ResourceViewer.vue
```diff
@@ -4,7 +4,7 @@
     <div class="resource-viewer-contents">
       <resource-viewer-block
         v-for="block in activeFlow?.blocks ?? []"
-        v-show="visibleBlocks.includes(block.uuid)"
+        v-show="visibleBlocks === null || visibleBlocks.includes(block.uuid)"
```
safydy-r, line 7: `visibleBlocks` is used on voto5, but here, we just display everything

src/store/undoRedo/index.ts
```diff
@@ -14,6 +15,7 @@ export interface ISnapshotModules {
 
 export interface ISnapshot {
   modules: ISnapshotModules,
+  sourcePage?: string,
```
safydy-r, line 18: I take the page id instead of the whole URL

object-required (approved):
Looks good, thank you :100:

bulatgab (approved):

PR #473, opened 2023-06-06
CORE-538, CORE-578 Fix getChangedKeys

object-required:
[Screencast from 2023-06-06 03:34:42 PM.webm](https://github.com/FLOIP/flow-builder/assets/3971957/0a8e9e59-fc6c-44b6-9399-02b9ca8bda11)

### DIFF
--- a/src/store/undoRedo/getChangedKeys.ts
+++ b/src/store/undoRedo/getChangedKeys.ts
@@ -7,12 +7,23 @@ export function getChangedKeys(a: object, b: object): string[] {
       const aVal = (a as Record<string, unknown>)[key]
       const bVal = (b as Record<string, unknown>)[key]
 
-      if (typeof aVal === 'object' && typeof bVal === 'object' && aVal !== null && bVal !== null) {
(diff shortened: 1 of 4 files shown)


bulatgab (approved):

src/store/undoRedo/getChangedKeys.ts
```diff
@@ -7,12 +7,23 @@ export function getChangedKeys(a: object, b: object): string[] {
...
     })
 
   return changedKeys
 }
+
+function ensureObject(obj: any): object {
+  return obj !== null && obj !== undefined ? obj : {}
+}
+
+function isObjectOrUndefined(obj: any): boolean {
```
safydy-r, line 27: 👍

safydy-r: @object-required , thanks for creating such methods. I'd suggest we put them under utils folder, eg: `floip\flow-builder\src\utils\check-utils.ts` ? In fact, these would be very useful for other implementations. What do you think?
- [] move utils under the appropriate folder

safydy-r (approved):
Pre-Approved with a change request ;)

PR #474, opened 2023-06-07
feature/CORE-538/CORE-601/deep-link-undo-redo

bulatgab:
[overview]

core-601

### DIFF
--- a/src/store/undoRedo/index.ts
+++ b/src/store/undoRedo/index.ts
@@ -3,7 +3,7 @@ import {Location} from 'vue-router'
 import {IRootState} from '@/store'
 import {IFlowsState} from '@/store/flow'
 import structuredClone from '@ungap/structured-clone'
-import {difference, isEmpty, union} from 'lodash'
+import {difference, union} from 'lodash'
 import {ActionTree, GetterTree, Module, MutationTree} from 'vuex'
 import router from '@/router'
 import {getDeepLink} from '@/store/undoRedo/deeplink'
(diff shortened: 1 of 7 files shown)


src/store/undoRedo/index.ts
```diff
@@ -116,14 +120,32 @@ export const mutations: MutationTree<IUndoRedoState> = {
 
 export const actions: ActionTree<IUndoRedoState, IRootState> = {
   async handleStateChange({commit, getters, rootState}) {
-    // Take a snapshot of the current state to avoid mutating> it
+    const currentModules: ISnapshotModules = structuredClone({flows: getters.currentSnapshot?.modules.flows ?? {}})
+    const newModules: ISnapshotModules = structuredClone({flows: rootState.flow})
+    const changedKeys = getChangedKeys(currentModules, newModules)
```
bulatgab, line 125: @object-required
I had to move the `getChangedKeys` call to the top, so that it could be used for deep link generation.
But there is no `newSnapshot` available here (it needs the `routeName` & `routeParams` to be complete - vicious circle).
So I used `structuredClone` on modules directly instead of `pack` function.

@safydy-r suggested we use `pack` (and I agree). If you think this is worth doing and you know how - please suggest! :)

safydy-r: When thinking, I'm ok with this implementation as newSnapshot is not available yet

safydy-r (approved):
LGTM ! Thanks @bulatgab 

@object-required , I'm merging this for now to move it quickly to QA. But feel free to review and leave a comment especially if there is any concern

src/store/undoRedo/index.ts
object-required: I'm ok as well, we can refactor later if/when it bothers us :wink:

PR #476, opened 2023-06-08
Feature/CORE-614/allow flow to be savable after deleting blocks

safydy-r:
[See voto5 PR for screen capture]

### DIFF
--- a/src/views/InteractionDesigner.vue
+++ b/src/views/InteractionDesigner.vue
@@ -60,7 +60,7 @@ import ClipboardRoot from '@/components/interaction-designer/clipboard/Clipboard
 import {Route} from 'vue-router'
 import {IBlock, IFlow, ILanguage, IResource, IResources, SupportedMode} from '@floip/flow-runner'
 import {ErrorObject} from 'ajv'
-import {MutationPayload} from 'vuex'
+import {ActionPayload, MutationPayload} from 'vuex'
 import {IValidationStatus} from '@/store/validation'
 
 Component.registerHooks(['beforeRouteUpdate'])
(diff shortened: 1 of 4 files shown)


src/views/InteractionDesigner.vue
```diff
@@ -251,7 +251,10 @@ export class InteractionDesigner extends mixins(Lang, Routes) {
     (global as any).builder = this
 
     // Listen to Flow updates to maintain "hasFlowChanges" state
-    this.$store.subscribe(this.handleFlowChanges.bind(this))
+    // 1- Listen to mutations first
+    this.$store.subscribe(this.handleFlowChangesViaMutations.bind(this))
```
object-required, line 255: We created `flow_removeBlock` mutation to make it work with undo/redo feature; I'd like to suggest picking that change instead of creating another event handler to keep things simple.

https://github.com/FLOIP/flow-builder/pull/472/files#diff-eea673c657b7edcdbf00ab176fcc8ec0a84d21e8a8dfa9ff8a014c7510a144a0R151-R160

The other commit https://github.com/FLOIP/flow-builder/commit/464073a0bf10de40521cbfe4249e1eb2d7b920e4

object-required (approved):
Looks good overall :100: thanks!
I left a suggestion for your consideration :wink:

src/views/InteractionDesigner.vue
safydy-r: Shall we decline this PR and wait for undo/redo instead?

safydy-r: If we cannot cherry pick commits from the undo/redo, then just replicate the solution here. Would be great to not have this 2nd binding

PR #477, opened 2023-06-09
Feature/CORE-538/fix router when used in voto5

safydy-r:
[When we consume the integration branch we have this error:

Screenshot 2023-06-09 073748

Cause:
- we're importing the router multiple times https://github.com/FLOIP/flow-builder/pull/471/files#diff-87e88e1682eb391eeeda70d0c90102a9cf9aacc4b0e9f47f465492b57a841b99R8, which calls `Vue.use(VueRouter)` multiple times then


safydy-r:
Merging this to fix the integration PR & unblock QA

PR #481, opened 2023-06-14
CORE-549, CORE-629 Custom Cypress commands

object-required:


src/store/undoRedo/index.ts
```diff
@@ -165,7 +165,12 @@ export const actions: ActionTree<IUndoRedoState, IRootState> = {
 
     // ####### force patch ##############
     if (isNewSnapshotFromPersistenceAction) {
-      commit('patchSnapshot', newSnapshot)
+      if (getters.hasCurrentSnapshot as boolean) {
+        commit('patchSnapshot', newSnapshot)
+      } else {
+        // We cannot patch non-existing snapshots
+        commit('addSnapshot', newSnapshot)
```
object-required, line 172: This case comes from this condition; I wondered when this may happen?

https://github.com/FLOIP/flow-builder/blob/b48cfff3dafbca86a893786c293aff4cbdabf45f/src/store/undoRedo/index.ts#L135-L139

cypress/support/commands.ts
```diff
@@ -24,14 +24,107 @@
...
+
+Cypress.Commands.add('addBlock', (menuChoices: string[]) => {
+  for (const choice of menuChoices) {
+    cy.get('[data-cy="blocks--menu"]')
+    cy.contains('[data-cy="blocks--menu-item"]', choice).click()
+  }
+
+  return cy.get('[data-cy^="block--"]').last().then((block) => {
+    return cy.wrap(block.attr('data-cy')!.replace('block--', ''))
+  })
```
object-required, line 94: This relies on appending new blocks to the DOM, unsure how fragile this may be.

cypress/e2e/undo-redo/add-a-block.cy.js
```diff
@@ -0,0 +1,15 @@
+describe('adding blocks', () => {
+  beforeEach(() => {
```
safydy-r, line 2: @object-required , could we extra basic command outside undo-redo folder? In fact, they are not fully undoRedo, and we may need them for other features later

safydy-r: I know I created the test under it  `cypress/e2e/undo-redo/create-new-flow.cy.js` but I think we can take the opportunity to structure a bit those specs. Your thoughts?

cypress/e2e/undo-redo/create-new-flow.cy.js
```diff
@@ -4,35 +4,19 @@ describe('create flows', () => {
...
-      .find('input')
-      .type(interactionTimeout)
-
-    cy.get('[data-cy="languages--selector"]').click()
-    cy.contains('.multiselect__option', 'English').click();
+    flowsListState().should('have.length', 0)
 
-    flowModuleState().its('isCreated').should('equal', false)
-    flowsListState().its('[0].label').should('equal', flowName)
+    cy.createFlow({
```
safydy-r, line 13: 👍

cypress/e2e/undo-redo/create-new-flow.cy.js
```diff
@@ -4,35 +4,19 @@ describe('create flows', () => {
...
+      label: 'This should be a valid empty flow',
+      interactionTimeout: 500,
+      languages: ['English', 'French'],
+    })
 
-    cy.get('[data-cy="create--btn"]').click()
     flowModuleState().its('isCreated').should('equal', true)
+    flowsListState().its('[0].label').should('equal', 'This should be a valid empty flow')
```
safydy-r, line 20: - [ ] use constant for "This should be a valid empty flow"

cypress/e2e/undo-redo/save-a-flow.cy.js
```diff
@@ -0,0 +1,19 @@
+describe('saving a flow', () => {
```
safydy-r, line 1: Nice, thanks for adding this

cypress/e2e/undo-redo/undo-redo.cy.js
```diff
@@ -0,0 +1,27 @@
+describe('undo-redo', () => {
```
safydy-r, line 1: I think it would be more specific if we name this "block-creation.cy.js" because it's under `undo-redo` folder. Besides, we plan to structure a bit the e2e folder to extra non undoRedo feature to be outside

cypress/e2e/undo-redo/undo-redo.cy.js
```diff
@@ -0,0 +1,27 @@
...
+  beforeEach(() => {
+    cy.createFlow({
+      label: 'undo-redo test',
+    })
+  })
+
+  it('undo block creation', () => {
+    cy.addBlock(['Content', 'Message']).then((uuid) => {
+      cy.wait(1000)
```
safydy-r, line 10: I'd suggest we create a custom cmd called `safeWait()` to encapsulate the `cy.wait(1000)`, and later we can easily change the time if needed

cypress/support/commands.ts
```diff
@@ -24,14 +24,107 @@
...
+      undo(): Chainable<void>,
+      redo(): Chainable<void>,
+      save(): Chainable<void>,
+      // connectExitToBlock(exitUuid: string, targetBlockUuid: string): Chainable<void>,
+      // dragBlock(blockUuid: string): Chainable<void>,
+    }
+  }
+}
+
+export interface ICreateFlowOptions {
```
safydy-r, line 44: can we use `Partial<IFlow>` instead? Would that work?

src/components/interaction-designer/block/Block.vue
```diff
@@ -3,6 +3,7 @@
     :id="`block/${block.uuid}`"
     :ref="`block/${block.uuid}`"
     class="block"
+    :data-cy="`block--${block.uuid}`"
```
safydy-r, line 6: yes ! 👍

cypress/support/commands.ts
```diff
@@ -24,14 +24,107 @@
...
+    return cy.wrap(block.attr('data-cy')!.replace('block--', ''))
+  })
+})
+
+Cypress.Commands.add('undo', () => {
+  cy.get('[data-cy="undo--btn"]')
+    .as('undoBtn')
+    .should('not.have.attr', 'disabled')
+
+  cy.get('@undoBtn').click()
```
safydy-r, line 112: didn't know we can do this, hihi. Thanks

safydy-r (changes requested):
In general it looks good to me, nice job @object-required .
I just want to suggest we take the opportunity to structure the `e2e/` folder a bit, and keep only undoRedo test under the `undo-redo/`. The basic ones would be extracted. Folder structure eg:
- e2e:
   - undo-redo
       - block-creation.cy.js
   - blocks
       - add-a-block.cy.js
       - select-a-block.cy.js
   - flows
       - save-a-flow.cy.js
       - create-new-flow.cy.js

safydy-r:
and could you pls add @pascaleli as reviewer? As the plan is to involve QA on tests writing too.

cypress/e2e/undo-redo/undo-redo.cy.js
safydy-r: Another alternative: We can put the `wait()` within undo and redo commands themselves, what do you think?

cypress/support/commands.ts
safydy-r: could you pls leave this as a comment in the code, so dev could improve it on the fly in the future?

cypress/support/commands.ts
safydy-r: No, it's not the IFlow, it's the representation of the flow creation page

cypress/e2e/flow-builder/undo-redo/undo-redo.cy.js
```diff
@@ -0,0 +1,22 @@
...
+  it('undo adding a block', () => {
+    cy.addBlock(['Content', 'Message']).then((uuid) => {
+      cy.undo()
+      cy.get(`[data-cy="block--${uuid}"]`).should('not.exist')
+    })
+  })
+
+  it('redo adding a block', () => {
+    cy.addBlock(['Branching']).then((uuid) => {
+      cy.undo()
```
safydy-r, line 17: nice, we don't have the wait() anymore 👍

safydy-r (approved):
LGTM, I like the folder structure

pascaleli:
Thanks @object-required  
These are very useful and nicely written commands . 
Even though I will need some explanation as to these two lines in the "addBlock" command 🤣

`return cy.get('[data-cy^="block--"]').last().then((block) => {
    return cy.wrap(block.attr('data-cy')!.replace('block--', ''))`
    
    I'm sure we will have the chance to talk about it. aside that this is a good one 👍🏽👍🏽

object-required:
> `return cy.get('[data-cy^="block--"]').last().then((block) => { return cy.wrap(block.attr('data-cy')!.replace('block--', ''))`

@pascaleli that's a way to find the most recently added block, which should be the last element marked with `data-cy`; the `return` inside allows us to get the UUID of that block.

PR #483, opened 2023-06-14
Feature/core 602/prompt on unsaved changes

GrimFunction:


src/views/InteractionDesigner.vue
```diff
@@ -260,6 +260,15 @@ export class InteractionDesigner extends mixins(Lang, Routes) {
     this.setActiveMainComponent({mainComponent: this.mainComponent})
   }
 
+  beforeMount(): void {
+    window.addEventListener('beforeunload', event => {
+      if (this.isEditable && this.hasFlowChanges) {
+        event.preventDefault()
+        event.returnValue = ''
```
GrimFunction, line 267: A returnValue with a string needs to be provided here (cannot use null). Previously, this string was used in the prompt, but that feature has since been deprecated (security reasons). So, an empty string is provided instead.

object-required (approved):

bulatgab (changes requested):
Thanks for the changes and for adding a demo.
It's nice that JS/WebAPI have this scenario covered 🙂
Left a note about `removeEventListener`.

bulatgab (approved):

PR #486, opened 2023-06-15
CORE-549, CORE-637 Saving should not produce extra snapshots

object-required:


cypress/e2e/flow-builder/undo-redo/persistence.cy.js
```diff
@@ -0,0 +1,22 @@
...
+
+    cy.addBlock(['Content', 'Message']).then(() => {
+      cy.save()
+      cy.get('[data-cy="save--btn"]').should('have.attr', 'disabled')
+
+      // Has one snapshot
+      cy.undo()
+      cy.get('[data-cy="undo--btn"]').should('have.attr', 'disabled')
+    })
+  })
```
object-required, line 21: Essentially, the change -- based off another branch, so the diff is not clean.

safydy-r (approved):

PR #488, opened 2023-06-15
Feature/core 549/core 653/test common block configs to mutate state

safydy-r:
If we agree on this approach for common configs, then we will go ahead to replicate the approach for all specific block configs for each block type.
core-653

What are covered: common block configs having UI
What are uncovered: if the config doesn't have an UI (eg: block.type)
Caveat: New block config (in future spec change) is not enforced, we should cover that in a code convention I think.

It looks pretty heavy with all block types, but with this tests we will be sure every configs are mutating the state. I've checked an alternative to use kind of linting to do the enforcement but I couldn't find a solution.


cypress/e2e/undo-redo/block-configs-should-update-flow-state/common-block-configs.cy.js
```diff
@@ -0,0 +1,149 @@
+/**
+ * In general, e2e should focus on what are VISIBLE from users perspective.
+ * But, as we have a bunch of configs to test, we can assume that:
+ * - if a block config UI updates the flow state, it should work with the UNDO/REDO feature
```
safydy-r, line 4: this is the main reason why we test the flow state instead of doing a long test like this "make a change on a config UI > assert something > undo > assert something > redo > assert something "

safydy-r: I will improve the comment by providing some clarification about "Vertical e2e test" here

pascaleli:
Thanks @safydy-r 
This is really cool 
with this we can write the configuration for all other blocks.

pascaleli (approved):

cypress/e2e/undo-redo/block-configs-should-update-flow-state/common-block-configs.cy.js
```diff
@@ -0,0 +1,149 @@
...
+    const flowsListState = () => flowModuleState().its('flows')
+
+    cy.addBlock(['Content', 'Message'])
+    const firstBlockState = () => flowsListState().its('[0].blocks[0]')
+    const firstResourcesState = () => flowsListState().its('[0].resources[0]')
+
+    // ####### Block label & name
+    cy.get('[data-cy="label--editor"]')
+      .find('textarea')
+      .type(String(blockConfigs.label))
```
safydy-r, line 50: Will check if we do need to cast with `String()`, I will remove if not needed. (applied for many typing like this)

cypress/e2e/undo-redo/block-configs-should-update-flow-state/common-block-configs.cy.js
```diff
@@ -0,0 +1,149 @@
...
+
+    cy.get('[data-cy="SMS-resource-variant-text--editor"]')
+      .find('textarea')
+      .type(String(blockConfigs.resources.sms))
+
+    cy.get('[data-cy="USSD-resource-variant-text--editor"]')
+      .find('textarea')
+      .type(String(blockConfigs.resources.ussd))
+    firstResourcesState().its('values').should((values) => {
+      console.debug('test rs', values)
```
safydy-r, line 72: will clean console debug

PR #490, opened 2023-06-16
Feature/core 538/core 631/e2e add remove connection

bulatgab:


cypress/.eslintrc.json
```diff
@@ -9,7 +9,7 @@
     "plugin:cypress/recommended"
   ],
   "rules" : {
-    "import/no-extraneous-dependencies": "off",
+    "import/no-extraneous-dependencies": ["error", {"devDependencies": true}],
```
bulatgab, line 12: `"off"` was too loose and didn't check for missing dependencies

package.json
```diff
@@ -121,6 +122,7 @@
     "babel-loader": "^9.1.2",
     "cookie-parser": "^1.4.5",
     "cypress": "^12.0.0",
+    "cypress-real-events": "^1.8.1",
```
bulatgab, line 125: I couldn't make `@4tw/cypress-drag-drop` library or cypress's `trigger` to work with our block connections.
Possibly, because the `<plain-draggable>` inside of the block exits is conditionally rendered on mouse hover.
And we can't simulate hovering with `@4tw/cypress-drag-drop` or via cypress's `trigger('mouseover')`.
I might be wrong though.

bulatgab: Unfortunately, `cypress-real-events` works only in Chromium-based browsers.
This means we won't be able to run cypress tests in Firefox, etc.
Please confirm that it's okay.

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
```diff
@@ -0,0 +1,41 @@
+describe('Creating a connection', () => {
```
safydy-r, line 1: - [ ] it's not only about "Creating". Let's rename it to "Creating/removing connections"

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    cy.addBlock(['Content', 'Message']).then(blockId => {
+      cy.get(`[data-cy="block--${blockId}"]`).as('firstBlock')
+    })
+
+    cy.addBlock(['Content', 'Numeric Response']).then(blockId => {
+      cy.get(`[data-cy="block--${blockId}"]`).as('secondBlock')
+    })
+
+    // wait for flow validation after adding blocks
```
safydy-r, line 15: what if we don't add this? Could you add that in the comment for more context pls?

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
```diff
@@ -0,0 +1,41 @@
...
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(300)
+  })
+
+  it('creates and removes a connection', () => {
+    cy.get('@firstBlock').find('[data-cy^="exit--"]').as('exit')
```
safydy-r, line 21: 1st time I see this ^ modifier, nice!

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
```diff
@@ -0,0 +1,41 @@
...
+    cy.get('body>.leader-line:last-of-type').as('connection')
+
+    cy.get('@connection').should('be.visible')
+
+    // remove connection
+    cy.get('@exit')
+      .realHover()
+      .click()
+
+    cy.get('@connection').should('not.exist')
```
safydy-r, line 39: @bulatgab , thanks for providing this test for the connections. I think, the remaining tests is on the undo/redo: Making sure we can undo and redo connections actions

- [ ] address this

package.json
safydy-r: No concerns about limiting with Chromium-based browsers only. We just need to inform QA about this.

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
safydy-r: cc: @object-required , @pascaleli in case you too

cypress.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
     setupNodeEvents(on, config) {
       // implement node event listeners here
     },
-    specPattern: 'cypress/e2e/**/*.cy.js',
+    specPattern: 'cypress/e2e/**/*.cy.{js,ts}',
```
object-required, line 8: :+1:

object-required (approved):

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
safydy-r: my suggestion is just to add few test after this line
```
cy.undo() // to undo the remove connection
// then assert something

cy.undo() // to undo the add connection with drag-n-drop
// assert something

cy.redo() // put back the connection
// then assert something

cy.redo() // remove back the connection
// then assert something
```

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
bulatgab: I think I saw it in @object-required's PR :) 
https://github.com/FLOIP/flow-builder/pull/481/files#diff-6c107e2390c5ad84e123f3be09d7b91ba35de8f8cc550d1f358c75806df47645R100

cypress/e2e/flow-builder/connections/creating-removing.cy.ts
bulatgab: This was to wait for the flow validation panel to disappear because it was overlapping the new block.
But with the block centering fixed it's not longer necessary.
Removed now.

src/store/builder/index.ts
```diff
@@ -361,33 +361,42 @@ export function generateConnectionLayoutKeyFor(source: IBlock, target: IBlock):
   ]
 }
```
bulatgab, line 363: FYI, in this file I've tried fixing the position of the newly added blocks on the canvas.
@object-required @safydy

safydy (approved):
Nice one, thnx

PR #491, opened 2023-06-20
Feature/core 549/core 630/cover add delete duplicates blocks

safydy-r:
Note: The deletion action has a bug, I created a ticket to address it https://viamoinc.atlassian.net/browse/CORE-664

For the other actions, the e2e assertions are all SUCCESS


cypress.config.ts
```diff
@@ -7,5 +7,7 @@ export default defineConfig({
     },
     specPattern: 'cypress/e2e/**/*.cy.js',
     experimentalRunAllSpecs: true,
+    "viewportWidth": 1440,
```
safydy-r, line 10: increased the default size of the view port

cypress/e2e/flow-builder/undo-redo/actions-on-block.cy.js
```diff
@@ -0,0 +1,55 @@
...
+    cy.addBlock(['Content', 'Message']).then((blockUuid) => {
+      messageBlockUuid = blockUuid
+      cy.get(`[data-cy="block--${blockUuid}"]`).should('exist')
+      cy.undo()
+      cy.get(`[data-cy="block--${blockUuid}"]`).should('not.exist')
+      cy.redo()
+      cy.get(`[data-cy="block--${blockUuid}"]`).should('exist')
+
+      // #### Duplicate that block
+      let duplicatedMessageBlockUuid = null
```
safydy-r, line 20: had to make nested `then()` because the custom commands are async instructions

cypress/support/commands.ts
```diff
@@ -88,17 +88,28 @@ Cypress.Commands.add('createFlow', (options: Partial<ICreateFlowOptions>) => {
...
 
-  // This relies on appending new blocks to the DOM, unsure how fragile this may be
-  return cy.get('[data-cy^="block--"]').last().then((block) => {
-    return cy.wrap(block.attr('data-cy')!.replace('block--', ''))
+      // Let the block's creation be registered in a snapshot
+      cy.wait(UNDO_REDO_WAIT_MS)
+    }
+
+    // Double check the blocks were created to insure the return of this custom command would be correct
+    flowsListState().its('0.blocks').then((blocks_2) => {
```
safydy-r, line 106: improved the previous fragile testing

object-required (approved):
LGTM :100: 

I'd like to suggest comparing lists of blocks instead of using indexes
https://github.com/FLOIP/flow-builder/pull/493

PR #495, opened 2023-06-22
feature/CORE-538/CORE-549/CORE-632/move-block

bulatgab:


package.json
```diff
@@ -91,6 +91,7 @@
     "vuex-class": "^0.3.2"
   },
   "devDependencies": {
+    "@4tw/cypress-drag-drop": "^2.2.4",
```
bulatgab, line 94: this time, on the contrary, the `cypress-real-events` didn't work but the `@4tw/cypress-drag-drop` did

bulatgab: So maybe we can use the first one for drag-n-drops, and the second one for arbitrary-delta move, or vice-versa (whichever works for the particular case)

cypress/e2e/flow-builder/blocks/moving.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+    })
+
+    cy.addBlock(['Content', 'Message']).then(blockId => {
+      cy.get(`[data-cy="block--${blockId}"]`)
+        .find('[data-cy^="block-handle--"]')
+        .as('blockArea')
+    })
+
+    // wait for flow validation to succeed and flow validation toolbar to disappear
+    // to avoid interference with block coordinates
```
bulatgab, line 17: And this time we need to `wait` because the flow validation might end in the middle of block move and mess up the coordinates.

cypress/e2e/flow-builder/blocks/moving.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+            cy.log('new metadata x,y:', newMetadataX, newMetadataY)
+
+            cy.wrap(newMetadataX - oldMetadataX).should('eq', deltaX)
+            cy.wrap(newMetadataY - oldMetadataY).should('eq', deltaY)
+          })
+        })
+      })
+    })
+  })
+})
```
bulatgab, line 50: I was able to extract the `getBlockVisualCoordinates` and `getBlockMetadataCoordinates` for reusability.
But mostly for readability (it was aweful before).

It's still nested and I'm happy with it.
But any ideas on how to improve it further are very welcome.

cypress/e2e/flow-builder/blocks/moving.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(200)
+  })
+
+  it('moves the block', () => {
+    const deltaX = 200
+    const deltaY = 300
+
+    // eslint-disable-next-line arrow-body-style
+    return getBlockVisualCoordinates().then(({x: oldVisualX, y: oldVisualY}) => {
```
safydy-r, line 27: I'm wondering if we always need these `return`  on async methods, I didn't use them in mine and it's working https://github.com/FLOIP/flow-builder/pull/494/files#diff-6e06e9c0a89fc89cc254694808341451273525e8839c678f5136bdedd7a4eaacR11

Could you pls try to remove them?

package.json
safydy-r: hum, that's interesting. Could you explain why do you say it didn't work?
I'd like to suggest we leave a note under an `.md` file in `e2e/` folder about this to guide test writer when to use cypress-drag-drop and when the cypress-real-events if we wanna keep both

cypress/e2e/flow-builder/blocks/moving.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+      return getBlockMetadataCoordinates().then(({x: oldMetadataX, y: oldMetadataY}) => {
+        cy.get('@blockArea').move({deltaX, deltaY})
+
+        // eslint-disable-next-line arrow-body-style
+        return getBlockVisualCoordinates().then(({x: newVisualX, y: newVisualY}) => {
+          return getBlockMetadataCoordinates().then(({x: newMetadataX, y: newMetadataY}) => {
+            cy.log('old visual x,y:', oldVisualX, oldVisualY)
```
safydy-r, line 34: didn't know `cy.log` was a thing :D

cypress/e2e/flow-builder/blocks/moving.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+            cy.log('new visual x,y:', newVisualX, newVisualY)
+
+            cy.wrap(newVisualX - oldVisualX).should('eq', deltaX)
```
safydy-r, line 37: any particular idea why do you propose to assert the visual coordinates & the metadata coordinates? No major concern, just curious.

object-required (approved):

cypress/e2e/flow-builder/blocks/moving.cy.ts
```diff
@@ -0,0 +1,74 @@
...
+            cy.wrap(newVisualY - oldVisualY).should('eq', deltaY)
+
+            cy.log('old metadata x,y:', oldMetadataX, oldMetadataY)
+            cy.log('new metadata x,y:', newMetadataX, newMetadataY)
+
+            cy.wrap(newMetadataX - oldMetadataX).should('eq', deltaX)
+            cy.wrap(newMetadataY - oldMetadataY).should('eq', deltaY)
```
safydy-r, line 44: My main concern is just on the Undo/Redo test. The same remark as I provided from the "add connection" test https://github.com/FLOIP/flow-builder/pull/490/files#r1232955027

Reminder, the whole ticket core-538 is to provide testing for Undo/Redo. We need to test those actions by calling `cy.undo() / cy.redo()` unless on subtasks about "mutating state"

safydy-r (changes requested):
thanks @bulatgab , left 02 requests:
- add a guide about when to use the appropriate drag-n-drop lib
- add undo/redo tests. Eg: if I undo the move, the block should come back from its original position. If I redo that, it takes the last position

cypress/e2e/flow-builder/blocks/moving.cy.ts
safydy-r: I'm good with them for now

cypress/e2e/flow-builder/blocks/moving.cy.ts
bulatgab: Ah, yes, of course. Will do

cypress/e2e/flow-builder/blocks/moving.cy.ts
bulatgab: At first, I was trying to do the visual testing - checking the `left` and `top` of the block - but it didn't work.
So I switched to Vuex instead. It worked.
And then I realized that I was doing something wrong with visual testing and fixed it.
So I have ended up with two working approaches, and I just kept them both.

cypress/e2e/flow-builder/blocks/moving.cy.ts
bulatgab: Yes, I can try 👍

package.json
bulatgab: I have added `cypress/e2e/cypress-guidelines.md`. Does it look okay?

cypress/e2e/flow-builder/blocks/moving.cy.ts
bulatgab: It worked!

cypress/e2e/flow-builder/blocks/moving.cy.ts
bulatgab: Done

cypress/.eslintrc.json
```diff
@@ -10,6 +10,8 @@
   ],
   "rules" : {
     "import/no-extraneous-dependencies": ["error", {"devDependencies": true}],
+    "jest/expect-expect": "off",
+    "jest/valid-expect-in-promise": "off",
```
bulatgab, line 14: This turns off the complaints about `it` (that there are not `expect` statements inside) and about returning promises.
Both are caused by Jest.

In my understanding, Cypress is using Chai under the hood and Jest is not needed nor used in our Cypress tests.
So we can turn the Jest rules off in the `cypress` folder (right?).
I couldn't find a one-line way though.
@safydy-r, @object-required

safydy-r: no concern about this for now, thanks

cypress/e2e/cypress-guidelines.md
```diff
@@ -0,0 +1,28 @@
+# Guidelines for writing Cypress tests
```
safydy-r, line 1: Thanks @bulatgab , this is a great summary & guidelines.

cc: @object-required , @pascaleli as this is important if you work on mouse actions

cypress/e2e/flow-builder/blocks/moving.cy.ts
```diff
@@ -0,0 +1,94 @@
...
+
+            cy.undo()
```
safydy-r, line 44: 👍

safydy-r (approved):
Awesome, thanks for making the change @bulatgab . 
I like how you clarified the mouse action libraries in cypress/e2e/cypress-guidelines.md

PR #498, opened 2023-06-22
Feature/core-549/core-660/test configs for content blocks

safydy-r:


cypress/e2e/flow-builder/blocks/adding.cy.js
```diff
@@ -10,6 +10,6 @@ describe('Adding blocks', () => {
   })
 
   it('from a sub-menu', () => {
-    cy.addBlock(['Content', 'Message']).should('match', /[\w\d]{8}-[\w\d]{4}-[\w\d]{4}-[\w\d]{4}-[\w\d]{12}/)
```
safydy-r, line 13: The way `addBlock()` doesn't need the 1st menu to be specified, otherwise it will try to click twice the "Set **Contact** Property" menu if we provide `addBlock(['Contact', 'Set Contact Property'])`. In fact, "Set Contact Property" contains the string "Contact"

So I took the opportunity to fix these block creation to avoid confusion in the future.

bulatgab (approved):

object-required (approved):

PR #500, opened 2023-06-23
CORE-661 Branching block config changes

object-required:


safydy-r (approved):
LGTM

cypress/e2e/undo-redo/block-configs-should-update-flow-state/branching-block-configs.cy.js
```diff
@@ -0,0 +1,48 @@
...
+    })
+
+    cy.addBlock(['Branching']).as('branchingBlock')
+
+    cy.get('[data-cy="add-exit--btn"]').click()
+    cy.get('[data-cy="advanced-exit-name--input"]').type(EXIT_NAME)
+    cy.get('[data-cy="advanced-exit-test-expression--input"]').type(EXIT_EXPRESSION)
+  })
+
+  it('should add choices', () => {
```
bulatgab, line 20: would it be better to call this 'exits' rather than 'choices'?
(and the same for another `it`?)

bulatgab (commented):
👀, 👍

PR #505, opened 2023-06-28
feature/CORE-591/rename-duplicated-blocks

bulatgab:
See demos in the main PR: https://bitbucket.org/voto/voto5/pull-requests/3412


src/store/flow/block.ts
```diff
@@ -180,7 +180,7 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
 
   block_setLabel({commit, dispatch}, {blockId, value}: {blockId: string, value: string | undefined}) {
     commit('block_setLabel', {blockId, value})
-    dispatch('block_setName', {blockId, value: snakeCaseOnSpaces(value as string)})
+    dispatch('block_setName', {blockId, value: generateBlockCodeFromLabel(value ?? '')})
```
bulatgab, line 183: The label can be undefined.
But it is not necessary to set the name as `undefined` because even an empty string would trigger validation error.
(And we are creating blocks with `name: ''`)

src/store/flow/block.ts
```diff
@@ -180,7 +180,7 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
 
   block_setLabel({commit, dispatch}, {blockId, value}: {blockId: string, value: string | undefined}) {
     commit('block_setLabel', {blockId, value})
-    dispatch('block_setName', {blockId, value: snakeCaseOnSpaces(value as string)})
+    dispatch('block_setName', {blockId, value: generateBlockCodeFromLabel(value)})
```
bulatgab, line 183: While the block label can be undefined, it is not necessary to set the name as undefined because even an empty string would trigger a validation error.
(And we are creating blocks with name: '')

src/store/flow/utils/blockLabelHelpers.ts
```diff
@@ -0,0 +1,32 @@
+import {IBlock} from '@floip/flow-runner'
+
+/**
+ * @example
+ * 'AAA' -> '(Copy) AAA' -> '(Copy 2) AAA', etc.
+ */
+export function incrementBlockLabelCopyNumber(oldLabel: string): string {
+  const FIRST_COPY_REGEX = /^\(Copy\)/
+  const NUMBERED_COPY_REGEX = /^\(Copy (\d+)\)/
```
object-required, line 9: Clever! :100:

object-required (approved):
Pre-approving with a minor test case change.

PR #506, opened 2023-06-30
feature/CORE-676/wrap-long-words-in-group-names

bulatgab:
Related voto5 PR: https://bitbucket.org/voto/voto5/pull-requests/3408

Demo in community builder:
Screenshot 2023-06-30 at 4 23 00 PM


src/css/customized/vue-multiselect.css
```diff
@@ -168,9 +168,8 @@ fieldset[disabled] .multiselect {
   background: #8C215C;
   margin-bottom: 5px;
   white-space: normal;
-  overflow: hidden;
   max-width: 100%;
-  text-overflow: ellipsis;
+  overflow-wrap: anywhere;
```
bulatgab, line 172: I've removed 
```
overflow: hidden;
text-overflow: ellipsis;
```
Hope that's okay.

object-required: I'm curious how the wrapping is handled in the design system components, but I think it's OK to have this change to unblock the release testing.

object-required (approved):
LGTM 💯

src/css/customized/vue-multiselect.css
bulatgab: 🤔 I'm not sure I understand, let's discuss!
We are not using the design system components with our selectors.
Do you mean that we should update the selectors in `viamo-ui` in the same fashion as in this PR?

bulatgab:
Merging this into master, although not QA'd directly.
The changes are similar to the ones in voto5 which will be tested. And if there is anything reported by the QA there, we can revisit this PR as well.

PR #509, opened 2023-07-31
feature/CORE-715/fix-choices-when-adding-channel

bulatgab:
see the demo in https://bitbucket.org/voto/voto5/pull-requests/3496


src/store/flow/flow.ts
```diff
@@ -496,33 +498,30 @@ export const actions: ActionTree<IFlowsState, IRootState> = {
     })
 
     // add missing variants
```
bulatgab, line 500: 👇 extract `resources.filter(...)` as `nonChoiceResources` and reuse it. No other changes

safydy (approved):

lakhbawa (approved):

PR #510, opened 2023-08-17
feature/CORE-820/audio-player-not-refreshing-upon-block-switch

bulatgab:
Child components in the block editor might have a setup logic in their `mounted` lifecycle methods.
When switching between blocks of the same type, the block editor stays open and gets reused, so the `mounted` logic of the children does not get called.
Adding a `key` prop with block.uuid should force re-rendering of the block editor in this scenario.

### DIFF
--- a/src/components/interaction-designer/block-editors/BlockEditor.vue
+++ b/src/components/interaction-designer/block-editors/BlockEditor.vue
@@ -3,15 +3,15 @@
     class="block-editor"
     :style="{top: topPosition}">
     <div v-if="activeBlock">
-      <!-- use the key prop to force block editor re-rendering -->
       <div
-        :key="activeBlock.uuid"
         class="tree-sidebar-edit-block"
-        :data-block-type="activeBlock.type"
-        :data-for-block-id="activeBlock.uuid">
+        :data-block-type="activeBlock && activeBlock.type"
+        :data-for-block-id="activeBlock && activeBlock.uuid">
+        <!-- use :key to re-render editor when switching blocks -->
         <component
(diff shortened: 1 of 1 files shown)


src/components/interaction-designer/block-editors/BlockEditor.vue
```diff
@@ -1,10 +1,14 @@
 <template>
-  <div class="block-editor" :style="{top: topPosition}">
+  <div
+    class="block-editor"
+    :style="{top: topPosition}">
```
bulatgab, line 4: ☝️ _eslint fixes_

src/components/interaction-designer/block-editors/BlockEditor.vue
```diff
@@ -1,13 +1,17 @@
...
     <div v-if="activeBlock">
       <div
         class="tree-sidebar-edit-block"
         :data-block-type="activeBlock && activeBlock.type"
         :data-for-block-id="activeBlock && activeBlock.uuid">
+        <!-- use :key to re-render editor when switching blocks -->
         <component
           :is="`Flow${activeBlock.type.replace('.', '')}`"
           v-if="activeBlock"
+          :key="activeBlock.uuid"
```
safydy, line 14: neat solution

safydy (approved):
LGTM

laserpants (approved):

src/components/interaction-designer/block-editors/BlockEditor.vue
bulatgab: found on Stackoverflow 😄

jcbashdown (approved):
Thanks Bulat!

